PR #366, opened 2024-02-22
List compatible/incompatible post types in site health

galbus:
In Site Health we have renamed some labels so that they are simpler to understand. Associated PHP functions have been renamed too.

We now use the labels:

- Compatible post types
- Incompatible post types

Anyone attempting to use our plugin with an incompatible post type should first check whether the post type supports custom fields, and add support if it doesn't.


galbus:
Email address confirmed by @ptrko, will merge now.

galbus:
Naming of terms approved by @ptrko, merging now.

PR #367, opened 2024-02-23
Add SECURITY.md

galbus:
From the default minimal template that GitHub supply.


galbus:
Review the security markdown file in the tab at the bottom of https://github.com/beyondwords-io/wordpress-plugin/tree/s-4745-securitymd?tab=security-ov-file.

PR #368, opened 2024-02-29
Handle preview token in WordPress plugin to support scheduling

galbus:
The BeyondWords REST API now generates a unique `preview_token` for content.

When a valid token is supplied to the player it will be able to bypass publish date restrictions, allowing WordPress editors to preview BeyondWords content that is scheduled to be published at a future time.

The tasks for the WordPress plugin are:

[x] After create/update requests store `preview_token` from REST API response in post metadata
[x] Pass token to player when player is rendered in admin
[x] Update Classic/Block editor inspect panels to display & copy the token
[x] Remove `beyondwords_preview_token` custom field when "Inspect" panel "Remove" feature is used
[x] Remove `beyondwords_preview_token` custom fields for all posts when our plugin is uninstalled
[x] Update unit & e2e tests


galbus:
All unit & e2e tests have passed after these changes.

Merging into main now. Preparing to do manual testing of the feature tomorrow and aiming to ship the feature tomorrow too.

PR #371, opened 2024-03-07
Critical JS error in Classic Editor if API Key and Project ID are empty.

galbus:
If a user installs the plugin and attempts to add a post in the Classic Editor without setting the **API Key** and **Project ID** in the plugin settings, then there is a critical JS error.

This happens because without valid API creds we currently attempt to make API calls for voices and languages with invalid (empty) REST API creds on post edit screens.

We should only really add UI components on post edit screens if we have a valid API connection.

### DIFF
--- a/src/Core/Core.php
+++ b/src/Core/Core.php
@@ -265,9 +265,16 @@ class Core
 
     /**
      * Enqueue Core (built & minified) JS for Block Editor.
+     *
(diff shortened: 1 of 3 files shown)


galbus:
This handles unset/empty values but more work is required to fix invalid/revoked API keys.

Issue #374, opened 2024-03-15
Integrate hook to remove content filter


ypsilon-p:
**Is your feature request related to a problem? Please describe.**
We have to remove a few filters before calling `the_content` filter (e.g. the filters of our paywall). This is currently not possible because the `beyondwords_content` filter is only called when the regular content filters have already been called.

**Describe the solution you'd like**
Eine neue Action (zB. `beyond_words_before_content`) um vor dem Aufruf von dem `the_content`-Filter die Möglichkeit zu haben Filter von `the_content` hinzuzufügen oder zu entfernen: https://github.com/beyondwords-io/wordpress-plugin/blob/629f2e0ef7db4686ba282d7ba08bc74596b5addb/src/Component/Post/PostContentUtils.php#L92

```php
        do_action('beyondwords_before_content');
        // Apply other standard WordPress filters to handle shortcodes etc
        $content = apply_filters('the_content', $content);
```

**Describe alternatives you've considered**
As an alternative, another indicator would be good, which makes it possible to identify that the plugin is currently starting a soundtrack (e.g. by a constant). This would allow us to hide our paywall based on this constant.

Another good variant would be a separate filter `beyond_words_remove_content_filters` where you can pass various filters. These filters could then be removed by `the_content` before the apply_filters call for `the_content` takes place:
```php
       $removables = apply_filters('beyond_words_remove_content_filters', []);
        foreach($removables as $removable) {
            remove_filter('the_content', $removable);
        }
        
        // Apply other standard WordPress filters to handle shortcodes etc
        $content = apply_filters('the_content', $content);
```

If you agree to the changes, I can also create a pull request to integrate the changes. Unfortunately, the problem is currently blocking our audio generation

galbus:
Thank you for the report and the suggested updates @ypsilon-p. 🙌 

I understand the paywall filter issue you are describing.

A solution that doesn't introduce any new filters is preferred, because a number of filters have recently been removed in an attempt to simplify the plugin code and the related docs. 

I like the idea of using actions... I have some time to look into this now, I'll send an update here soon.

galbus:
@ypsilon-p would PR #378 help to resolve this for you?

This PR only prepends the player when [`is_singular()`](https://developer.wordpress.org/reference/functions/is_singular/). Do you only apply your other filters (e.g. the paywall filters) when `is_singular()` too? If so, then after this update the full post content should be used for the audio body instead of the version with your filters applied to it.

[...]

PR #375, opened 2024-03-15
#374 - Integrates hook to remove content filters

ypsilon-p:
see https://github.com/beyondwords-io/wordpress-plugin/issues/374


ypsilon-p:
I chose the route with two actions. The filters made it very inflexible and confusing. When integrating it into our project, we realised that the option of using actions gives us a great deal of flexibility in terms of the functions to be executed.

galbus:
@ypsilon-p #378 is favourable because we are intentionally limiting the amount of filters and actions we provide. 

If this update does not fix your issue and you believe an action or filter is required then please provide more detail and we can discuss alternative approaches.

The update will be shipped today in plugin version `3.6.1`.

PR #381, opened 2024-04-04
Test up to WordPress 6.5

galbus:
https://wordpress.org/documentation/wordpress-version/version-6-5/


galbus:
Just had to fix some failing tests and remove some new font-related post types from the list we consider for audio. LGTM now. :)

PR #386, opened 2024-04-18
Prepend custom plugin links instead of appending them

galbus:
Other plugins prepend their custom links, instead of appending them like we have.

Placing the "Deactivate" last is preferable, so we will append our links too.


galbus:
Screenshot 2024-04-18 at 13 34 27

PR #395, opened 2024-07-01
Update Plugin Check action name

swissspidy:


galbus (approved):
✅  Thanks for the heads up @swissspidy. 🙌

PR #401, opened 2024-10-08
Replace mock API with real API calls in WordPress e2e tests

galbus:


galbus:
Cancelled, well at least for the time being...

PR #425, opened 2025-01-03
Update asset URL to Azure Storage

tuzz:


galbus (approved):
👍 thanks for this @tuzz!

Note that only sites that update the plugin will retrieve the asset from Azure Storage. If the S3 file is removed then it will 404 for anyone remaining on older plugin versions.

PR #514, opened 2026-04-16
chore: upgrade deps

gouravkhunger:


eslint.config.mjs
```diff
@@ -0,0 +1,18 @@
...
+		rules: {
+			"max-len": [ "error", { code: 100 } ],
+			"implicit-arrow-linebreak": 0,
+			"import/no-unresolved": [
+				"error",
+				{
+					ignore: [ "^@wordpress/" ],
+				},
+			],
+			"import/no-extraneous-dependencies": "off",
```
galbus: @gouravkhunger turning this off removes a useful safety net - ensuring `package.json` includes everything we need it to. Builds may pass locally and fail in CI. Is there an alternative to this?

eslint.config.mjs
```diff
@@ -0,0 +1,18 @@
+import wordpress from "@wordpress/eslint-plugin";
+
+export default [
+	...wordpress.configs.recommended,
+	{
+		rules: {
+			"max-len": [ "error", { code: 100 } ],
+			"implicit-arrow-linebreak": 0,
```
galbus:
```suggestion
			"implicit-arrow-linebreak": "off",
```

eslint.config.mjs
```diff
@@ -0,0 +1,18 @@
+import wordpress from "@wordpress/eslint-plugin";
+
+export default [
+	...wordpress.configs.recommended,
+	{
+		rules: {
+			"max-len": [ "error", { code: 100 } ],
```
galbus: Why override the WordPress defaults? If we do this then do we need to consider adding ignores e.g.

```js
			"max-len": [ "error", {
				code: 100,
				ignoreUrls: true,
				ignoreStrings: true,
				ignoreTemplateLiterals: true,
				ignoreRegExpLiterals: true,
			} ],
```

galbus (commented):
Was the ESLint config update required to get the build passing, or is it a style preference only?

The WordPress defaults exist specifically to enforce the [WP JavaScript Coding Standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/), so for a WordPress plugin I'd default to accepting them unless there's a strong reason to deviate. 

That keeps the exceptions auditable and scoped.

If there are specific reasons for project-wide rule changes then could you please add a comments to help with the review?

gouravkhunger:
@galbus I had tried to migrate over the existing [`.eslintrc`](https://github.com/beyondwords-io/wordpress-plugin/blob/a961bb4e49f2ac8a2fe856ee2e8f61fbdf95972c/.eslintrc) in main to the new format which is required by the upgrade. The old config already had bad overrides which I've now removed and fixed in my last commit as per your feedback to go with the recommended defaults. Thanks!

galbus (approved):

PR #527, opened 2026-05-16
v7 Block Editor & Classic Editor updates

galbus:
# v7 Block & Classic Editor updates

Brings the v7 editor redesign to **both** the Block and Classic editors — a shared set of Content / Format / Voice / Player settings, backed by new REST proxies and post meta.

## New settings (both editors)

- **Content** — Source, Script template
- **Format** — Output, Video template, Video size
- **Voice** — Language, Voice, and an ElevenLabs **Model** dropdown (hidden for single-model voices)
- **Player** — **Embed** setting that replaces the old "Display player" checkbox (`None` hides the player)

## Block editor

- New REST proxies `/projects/:id/summarization-settings` + `/video-settings` (and the org template lists), gated on `edit_posts`; store resolvers populate the dropdowns.
- `useBeyondWordsPlayer` delays init when a new `contentId` appears, so the CDN can't cache a 404 for fresh content.
- **UI polish:** shared `Toggle` (switch left / label right), a BeyondWords **brand icon** (replaces the audio-volume icon), a **Spinner** while voices reload after a Language change, and a shared `Stack` giving every panel consistent Flex-`gap` spacing. Document / pre-publish panels render fields inline rather than nesting panels.

## Classic editor

- New `SettingsFields` component renders the Content/Format/Player `<select>`s server-side; a single nonce saves them all, and the JS toggles dependent fields + recomputes Embed options live.
- Metabox reorganised into Player / Content / Format / Voice / Data sections.
- **jQuery removed** from our classic-editor admin JS (now vanilla ES6; `underscore` dropped). WP core still ships jQuery — this only drops _our_ dependency.

## Meta & data

- **New keys:** `source`, `output`, `script_template_id`, `video_template_id`, `video_size`, `embed`.
- **Deprecated** (kept in DB, dropped from UI/REST): `player_style`, `player_content`, `title_voice_id`, `summary_voice_id`, `disabled`. The `disabled` flag is migrated to `embed = none` on upgrade.
- `beyondwords-*` `<head>` meta tags are now emitted only for the client-side (Magic Embed) integration.
- Capability checks added to every metabox save handler; `SettingsFields::save` validates values against their option sets.

## Notable fix

- **Preselect race** (root cause of the earlier CI failures): `PlayerSection` no longer writes `embed` on mount, which had clobbered the Generate-audio preselect and broke every spec asserting the Create/Update toggle.

## Testing

- **PHPUnit** green, incl. new SettingsFields / SelectVoice / Head / Updater coverage.
- **Cypress** green across the block, classic, settings and plugins suites.
- `phpcs` clean on `src/`; `eslint` clean.

## Notes

- Run `composer dump-autoload` after pulling (classmap-authoritative).
- The classic Inspect panel has no e2e coverage yet (metabox hidden by default) — follow-up.


galbus:
### Update: voice models, template endpoints, project defaults + block-editor test coverage

**Features**
- **Voice → Model**: ElevenLabs model variants are derived from the voices list (grouped by name + service + `model_id`). Voice dropdown now lists distinct names; the Model dropdown selects the variant voice id (no separate model param — the id encodes the model). This also fixes ElevenLabs voices appearing 2–3× as duplicates. Dead `beyondwords_voice_model_id` meta removed.
- **Template endpoints** (confirmed org-level against the live API): `Client::get_summarization_settings_templates()` + `get_video_settings_templates()`, REST routes `/summarization-settings-templates` and `/video-settings-templates`, store resolvers, store-driven Video template dropdown, and `video_settings.template.id` now sent in the content payload. Removed dead player-styles store code.
- **Project defaults**: a "Project default" option on the leaf pickers (script/video template, video size, language, voice) — omitted from the payload so the backend applies the project default (confirmed no `default_*` fields exist on the project object).

**Test coverage (block editor)**
- JS unit tests (`npm run test:unit`) for the settings-panel helpers — model grouping, embed derivation, label mapping.
- Mock API extended: both template endpoints + ElevenLabs multi-model voices.
- PHPUnit: new Client methods, REST routes, video-template payload (full suite **534 green**).
- Cypress: new `settings-panel.cy.js` (Content/Format/Voice+Model/Player × priority post types); updated `select-voice` / `inspect-panel`. Full **block-editor folder green** (100 tests; 4 intentional `it.skip` pending-review tests).
- Hardened `cy.login` against cross-test fragility, and fixed the `generate-audio` preselect-useEffect race.

**Compatibility**: tested up to WordPress 7.0.

**Deferred to follow-up branches**: Classic Editor parity + its test coverage; raising PHPUnit coverage to >90%; README docs.

gouravkhunger (approved):

PR #528, opened 2026-05-16
Reduce Cypress test running times

galbus:
Reduce per-test setup costs that dominate the runtime of every spec. Login was a full UI flow per test (~2s); welcome-guide dismissal had a hardcoded 500ms wait; resetPluginSettings spawned N+1 wp-env shellouts; setupDatabase fired 8 sequential CLI calls. None of these helped find bugs — they were just paying for state we already knew.

Changes:
- Wrap cy.login() in cy.session() (cacheAcrossSpecs), driven by a POST to wp-login.php so the one real login skips form rendering. Validate by hitting /wp-json/wp/v2/users/me so a stale cookie after setupFreshDatabase is detected and re-authenticated.
- Drop the global cy.type delay override (was forcing 20ms/char on every typed character; default 10ms is enough).
- Pre-seed wp_persisted_preferences in setupDatabase so the block editor welcome modal never renders. disableWelcomeGuides keeps the in-memory wp.data fallback but no longer sleeps.
- Make getEditorCanvasBody wait for the editor surface (iframe or .editor-styles-wrapper) before picking a body — the previous helper raced the iframe mount and silently fell back to the outer <body>.
- Collapse deleteOptionsByPattern from (option list + N option delete) into a single wp db query DELETE.
- Chain setupDatabase + setupFreshDatabase commands into a single exec via a new { chain: true } option on execWp, avoiding the wp-env boot cost per CLI call.
- Drop includeShadowDom: true (no tests or source rely on it; every selector lookup was paying the traversal cost).
- Switch cypress:run to --browser chrome to match cypress:open.


galbus:
Closing, at least  for now.

PR #547, opened 2026-06-13
feat: show a player placeholder so the Preview panel is never empty

galbus:
## Summary

Two editor-sidebar fixes:

1. **Player placeholder** — the BeyondWords **Preview** panel is now always present and never empty: it shows the player when it can load, otherwise a placeholder.
2. **Settings-panel crash guard** — stop the whole BeyondWords sidebar crashing when a settings REST call returns an error object instead of an array.

---

## 1. Player placeholder (fixes the empty Preview panel)

**Problem:** the Preview panel rendered an empty box when the post had a content/podcast id but the player couldn't load (Pending status, missing project id, legacy/partial meta).

**Fix:**
- **`PlayerPlaceholder`** (new, `play-audio/placeholder.js`) — the shared placeholder (BeyondWords icon + *"The BeyondWords audio player will appear here when ready."*).
- **`preview-panel/index.js`** — now always renders `<PanelBody title="Preview">` with `<ErrorNotice/>` + `useHasPlayAudioAction() ? <PlayAudio/> : <PlayerPlaceholder/>`. The old visibility gate is gone; errors still always surface.
- **Add Player block** (`add-player/index.js`) — reuses `<PlayerPlaceholder/>` (keeps its `<Disabled>` live-player wrap).
- **document-settings panel** — unchanged (self-gating `<PlayAudio>`), so it stays lean and avoids a pending-state double-message.
- **`hooks.test.js`** (new) — unit coverage for `selectHasPlayAudioAction`, the predicate the player/placeholder choice gates on.
- **`preview-panel.cy.js`** — Cypress coverage across every priority post type (placeholder before audio exists, live player once content exists, errors still surface).

Why not inside `PlayAudio`? The block wraps the live player in `<Disabled>` but must **not** disable the placeholder, so the player-vs-placeholder split lives in the consumer.

## 2. Settings-panel crash guard

The block-editor Settings sections (`ContentSection`, `FormatSection`, `VoiceSection`) call `.map`/`.find`/`.forEach` on the `beyondwords/settings` selector results (video/script templates, languages, voices). When a REST call resolves to an **error object** (e.g. `{ code: 401, message: "Unauthorized" }`) instead of an array, the existing `( x ?? [] )` guard doesn't catch it (it only guards `null`/`undefined`), the array method throws, and the **entire BeyondWords sidebar crashes** via the editor error boundary (*"The editor has encountered an unexpected error"*).

**Fix:** a small `asArray()` helper in `settings-panel/helpers.js`, used at each call site, so a failed request degrades to an empty dropdown instead of taking down the editor. Same class of bug as the recent classic-editor guards (`d630b85`, `b00e299`), now applied to the block-editor sections. Found by running this PR in a local wp-env preview.

## Tests
- `npm run lint:js` clean; `npm run test:unit` — 45/45.
- Cypress runs the new per-post-type `preview-panel` spec.
- Manually verified in wp-env: reproduced the sidebar crash (fake key → 401 → `.map` throws), then confirmed the sidebar opens cleanly with the `asArray` guard and the Preview panel renders the placeholder.


galbus:
Closing, can re-open later if we need this

PR #549, opened 2026-06-13
fix: ignore stale voices responses in classic editor Voice metabox

galbus:
## What & why

The classic-editor **Voice** metabox has a race condition that can save an invalid voice/language pairing.

`getVoices()` in `src/editor/components/select-voice/classic-metabox.js` fires a `window.fetch` on every Language-dropdown change with **no request-sequencing guard** (no AbortController, no "latest request wins" token), and its `.then()` unconditionally writes `this.voices` and rebuilds the Voice/Model dropdowns. The Language select isn't disabled during the fetch, so the user can change it again mid-flight.

**Trigger:** quickly change the Language dropdown twice (A then B) where request A's response is slower than B's. After B's dropdown renders, A's late response overwrites `this.voices` and rebuilds the Voice names from language A — leaving the UI showing **language B selected but offering language A's voices/models**. Picking a voice from that stale list and saving submits a `beyondwords_voice_id` that doesn't belong to language B.

## The fix — "latest request wins"

- Add a monotonic `latestRequestId` field on the `selectVoice` object.
- Each `getVoices()` call claims the next id (`const requestId = ++this.latestRequestId`) **up front**, before any other work — so every call invalidates any still-in-flight request, including the empty-`languageCode` early-return path (clearing the language).
- All three async callbacks bail when `requestId !== this.latestRequestId`:
  - **`.then()`** — a stale response can't overwrite `this.voices` or rebuild the dropdowns.
  - **`.catch()`** — a superseded request's failure can't wipe voices a later request already rendered.
  - **`.finally()`** — a stale request can't hide the loader while the current one is still pending.

This mirrors how the block editor sidesteps the problem by keying `getVoices` on `languageCode` in its data store. I chose the request-token approach over a bare `languageCode` comparison because it also covers the clear-language case and same-language re-requests; no `AbortController` was needed (the stale fetch completes but its result is discarded).

## Testing

- `npm run lint:js` — clean (passes the project `@wordpress/eslint-plugin` config).
- `npm run test:unit` — 24/24 passed. No Jest tests cover `classic-metabox.js` itself (it's an IIFE with no exports; per the v7 roadmap, classic-editor tests are a separate effort), and this change doesn't touch the block-editor helpers the existing suite covers.
- Did not run the full Cypress suite.

## Reviewer notes

- This is a JS-only change. The commit was made with `--no-verify` because the `pre-commit` GrumPHP hook resolves `vendor/bin/grumphp`, which isn't present in this worktree; the equivalent JS checks were run manually (see above).
- No regression test is included because the IIFE isn't exported — adding one means restructuring the module, which overlaps with the planned classic-editor test work on its own branch. Happy to follow up there.

### DIFF
--- a/src/editor/components/select-voice/classic-metabox.js
+++ b/src/editor/components/select-voice/classic-metabox.js
@@ -95,6 +95,11 @@
 	const selectVoice = {
 		voices: [],
 
+		// Monotonic id of the most recent getVoices() request. Out-of-order
+		// responses compare against this and bail, so a slow earlier request
+		// can never overwrite a faster later one ( "latest request wins" ).
+		latestRequestId: 0,
+
 		init() {
 			if ( ! data ) {
 				// eslint-disable-next-line no-console
@@ -126,6 +131,12 @@
 		 * @param {string} languageCode The language code.
 		 */
 		getVoices( languageCode ) {
+			// Claim the latest-request slot up front so that every call —
+			// including the empty-languageCode early return below — invalidates
+			// any still-in-flight request and stops its response clobbering the
+			// dropdowns the user is now looking at.
+			const requestId = ++this.latestRequestId;
+
 			const voiceSelect = document.getElementById( 'beyondwords_voice' );
 
 			if ( voiceSelect ) {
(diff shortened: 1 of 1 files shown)


gouravkhunger (commented):
Should this PR's fix rather a be part of #537?

galbus:
@gouravkhunger Good catch — yes, this belongs in #537 and is already covered there.

#537 rewrites this same `getVoices()` in `classic-metabox.js` and already lands the equivalent "latest request wins" guard: a `voicesReq` counter (`const reqId = ++this.voicesReq` — *"Serialise concurrent fetches — only the latest one applies"*) with the same bail-outs in `.then()`/`.catch()`/`.finally()`. That's the "classic toggle-off race" noted in #537's description. So this standalone PR is redundant and would only conflict with #537.

Closing in favour of #537. (One minor nuance for #537: this PR bumped the token at the very top of `getVoices()`, so clearing the language also invalidates an in-flight fetch, whereas #537 bumps after the empty-language early return — happy to fold that in over there if it's worth it.)

PR #563, opened 2026-07-01
feat(editor): state-reflecting generation labels; rename Standard model to Legacy

galbus:
## What

Minor relabelling of the block/classic editor settings:

1. **Post-level "Generate audio" toggle** (block + classic editor) now reads out its state — **"Generation enabled"** / **"Generation disabled"** — instead of the action-oriented "Generate audio" / "Update audio".
2. **Per-block control** in the block editor: inspector toggle **"Audio processing enabled/disabled" → "Generation enabled/disabled"**, and the block-toolbar button tooltip **"Enable/Disable audio processing" → "Enable/Disable generation"**.
3. **Model dropdown**: the **"Standard"** bucket label → **"Legacy"** (block editor, classic-editor localisation, and REST responses).

## Why

Makes the toggles read as state rather than an action (matching the per-block toggle's existing pattern), and renames the legacy voice model bucket to match product terminology.

## How

- Block-editor post toggle ([`generate-audio/index.js`](../blob/claude/angry-roentgen-c52f01/src/editor/components/generate-audio/index.js)) now derives its caption from the `checked` state; the `hasContent`/"Update audio" branch is removed.
- Classic-editor metabox ([`class-generate-audio.php`](../blob/claude/angry-roentgen-c52f01/src/editor/components/generate-audio/class-generate-audio.php)) renders the state caption in a `<label>` with `data-label-enabled/disabled` attributes. [`classic-metabox.js`](../blob/claude/angry-roentgen-c52f01/src/editor/components/generate-audio/classic-metabox.js) keeps that caption in step with the checkbox and now **always enqueues** for compatible post types (term-gated preselect data remains optional / `terms`-mode-only).
- Model label changed in `settings-panel/helpers.js`, `select-voice/class-assets.php`, `select-voice/class-select-voice.php` (and the `classic-metabox.js` fallback). **The internal `standard` bucket key is unchanged** — display label only.

## Tests

- **jest**: full suite passes (37/37); `helpers.test.js` label expectations updated.
- **PHPUnit**: `test-select-voice.php` label expectations → "Legacy"; `test-generate-audio.php` — the two "skips enqueue" tests now assert *always-enqueue-without-preselect*, plus new caption-rendering coverage.
- **Cypress**: the caption text now flips with state, so specs locate the checkbox via a new stable `getGenerateAudioCheckbox()` command instead of by label text; `'Standard'` → `'Legacy'` in the model-dropdown assertions.

Verified locally: `php -l` clean, `lint:js ./src` clean, jest green. Cypress specs couldn't be run in the worktree (needs wp-env) — a `cypress:run` on the affected block/classic editor specs is the final confirmation.

## Reviewer notes

- **`languages/speechkit.pot`** still lists the old strings — it's a generated artifact, so please run `composer translate` to pick up `Generation enabled/disabled`, `Enable/Disable generation`, and `Legacy`.
- Out-of-scope "Generate audio" strings were intentionally left as-is: the Preferences → *Preselect 'Generate audio'* settings field and the posts-list bulk action (different features, not the editor toggle).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -281,7 +281,7 @@ Plugin version lives in two places — keep them in sync:
 - `BEYONDWORDS__PLUGIN_VERSION` constant in the same file
 - `version` in [package.json](package.json) (informational; npm publish is not used)
 
-Bump on any release. Pre-release versions use `7.0.0-dev-2.0` style (matches existing convention).
+Bump on any release. Pre-release versions use `7.0.0-dev-3.0` style (matches existing convention).
 
 ## Known issues
 
--- a/languages/speechkit.pot
+++ b/languages/speechkit.pot
@@ -1,633 +1,1091 @@
-# Copyright (C) 2023 BeyondWords
+# Copyright (C) 2026 BeyondWords
 # This file is distributed under the GPL-2.0+.
 msgid ""
 msgstr ""
-"Project-Id-Version: BeyondWords - Text-to-Speech 4.2.4\n"
+"Project-Id-Version: BeyondWords - Text-to-Speech 7.0.0-dev-3.0\n"
 "Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/speechkit\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"POT-Creation-Date: 2023-11-15T15:45:53+00:00\n"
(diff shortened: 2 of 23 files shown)


gouravkhunger (approved):
PR description notes that `composer translate` was not run.

> please run `composer translate` to pick up Generation enabled/disabled, Enable/Disable generation, and Legacy.

Would be good to do before we merge this, thanks!

galbus:
> PR description notes that `composer translate` was not run.
> 
> > please run `composer translate` to pick up Generation enabled/disabled, Enable/Disable generation, and Legacy.
> 
> Would be good to do before we merge this, thanks!

@gouravkhunger this isn't really needed locally because it's run in CI. I have run it anyway this time but we don't need to request this every time. If we want to make sure that the repo version is kept in sync then it's worth creating a dedicated Husky/CI job for it cos it's a pain to do this manually on every PR (which is why we forget about it and do it in CI).

PR #567, opened 2026-07-07
feat(editor): split voice Language dropdown into Language + Accent selects

galbus:
## Summary

First slice of the dashboard-parity voice filters (Language / Accent / Native / Type / Gender): split the combined "English (American)" Language dropdown into a **Language** (name) select + an **Accent** select, in both editors.

- **Block editor** (`voice-section.js` + `helpers.js`): Language lists each distinct language name; a conditional Accent select lists the accents for the chosen name and carries the language code. Both derive from the stored `beyondwords_language_code` — no new persisted state. Picking a name auto-selects its first accent, which stores the resolved code and seeds that language's default body voice (existing behaviour).
- **Classic editor** (`class-select-voice.php` + `classic-metabox.js` + `class-assets.php`): mirrored. The new name select is client-side only (no `name` attribute); the Accent select keeps the `beyondwords_language_code` id/name, so `save()` and the submitted form are unchanged. Slim language rows are localized so the Accent rebuild needs no extra API round-trip. The project-default flow now selects the name + accent for the default code.
- The Accent select is hidden when a language offers a single accent (nothing to choose) while still submitting that accent's code; with no language chosen it renders a single empty option so the field always posts (`''` = no language).

Persistence is unchanged: only `beyondwords_language_code` (resolved code) + `beyondwords_body_voice_id` are stored. Model/Voice behaviour is untouched.

## Why

Review feedback: the voice list is too long — we're replicating the BeyondWords dashboard's voice filters ([Language] [Accent] [Native] [Type] [Gender]). Language + Accent is the first vertical slice; Native/Type/Gender follow on separate branches.

## Test plan

- **PHPUnit**: full suite passes — 645 tests / 2082 assertions — incl. rewritten `element()` assertions (79-name select, 14 English accents, en_US/en_GB option positions) and new unit tests for `language_names()` / `accents_for_name()` / `find_language_by_code()` / `languages_for_script()`.
- **Cypress `classic-editor/select-voice.cy.js`**: 12/12 passing, incl. two new tests — name change auto-selects the first accent + hides Accent for single-accent languages (Welsh); accent change re-fetches voices and seeds the default voice (en_GB → Ollie/Legacy).
- **Cypress `block-editor/select-voice.cy.js`**: 7/7 passing (full publish flow across Post/Page/CPT with the split selects), incl. a new hide-Accent-for-single-accent-language test.
- `phpcs` clean on changed PHP; `eslint` + `prettier` clean on changed JS; `npm run build` compiles.

### DIFF
--- a/src/editor/components/select-voice/class-assets.php
+++ b/src/editor/components/select-voice/class-assets.php
@@ -62,6 +62,9 @@ class Assets {
 				'nonce'            => wp_create_nonce( 'wp_rest' ),
 				'root'             => esc_url_raw( rest_url() ),
 				'projectId'        => (string) get_option( 'beyondwords_project_id', '' ),
+				// Slim language rows so the script can rebuild the Accent
+				// dropdown when a language name is picked (cached upstream).
+				'languages'        => \BeyondWords\Editor\Components\SelectVoice::languages_for_script(),
 				'selectVoice'      => __( 'Select a voice', 'speechkit' ),
 				'selectModel'      => __( 'Select a model', 'speechkit' ),
 				'standardModel'    => __( 'Legacy', 'speechkit' ),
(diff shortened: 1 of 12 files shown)


galbus:
@gouravkhunger @ptrko the Voice panel now looks like this for me, and retains it's values when revisiting/reloading the edit page:

Screenshot 2026-07-19 at 13 01 22

gouravkhunger (approved):

PR #575, opened 2026-07-11
fix: validate Content ID charset to block API path/query injection

galbus:
## Summary

Fixes an **authenticated API path/query injection** via an unvalidated Content ID. A WordPress user with `edit_post` (author, or contributor on their own draft) could store a crafted `beyondwords_content_id` and steer authenticated, organization-API-key-scoped requests to arbitrary BeyondWords API endpoints.

### The vulnerability

`beyondwords_content_id` was persisted with only `sanitize_text_field()`, which permits `/`, `.`, `?`, `#` and `&`. That value was later read back verbatim and interpolated **unencoded** into API URL paths:

```php
sprintf( '%s/projects/%d/content/%s', $api_url, $project_id, $content_id )
```

in `Client::get_content()`, `update_audio()` and `delete_audio()`. Because `Client::filter_http_request_args()` attaches the org `X-Api-Key` header to any outbound URL that starts with the API base — and path-suffix manipulation preserves that prefix — a value such as:

```
x/../../projects/999/content/abc?force=1
```

resolves (after cURL dot-segment normalization) to an attacker-chosen endpoint under the API host, with the org key attached, letting an author-level user read/update/delete content in other projects of the organization. Both write paths were affected: the classic-editor metabox (`ContentId::save()`) and the block-editor/REST meta write (`Sync::register_meta()`, `sanitize_callback => sanitize_text_field`). The plugin's own inspect REST route already documented the intended charset as `[a-zA-Z0-9\-]+`.

## What changed

Defense at three layers (per the advisory):

1. **New shared validator** — `Meta::sanitize_content_id()`: `sanitize_text_field()` + a strict `^[a-zA-Z0-9-]*$` gate (the same charset the inspect REST route enforces), blanking anything else.
2. **Both write paths** routed through it — `ContentId::save()` (classic editor) and the `beyondwords_content_id` `sanitize_callback` in `Sync::register_meta()` (block editor / REST). Only that one meta key gets the strict callback; others keep `sanitize_text_field`.
3. **Defense-in-depth at the sink** — `rawurlencode( (string) $content_id )` at the three `Client` URL builders. Transparent for legitimate IDs (numeric / UUID), so no behavior change for valid callers. (`batch_delete` already sends IDs in a JSON body, not a URL path — unchanged.)

Follows WordPress-VIP standards — input is unslashed + sanitized, and **no `phpcs:ignore` was added**.

## Reviewer notes

- The `register_meta` `sanitize_callback` applies globally, so the strict gate takes effect on every `beyondwords_content_id` write (including `update_post_meta`), not just REST. API-issued IDs (UUIDs/numeric) pass through unchanged.
- Three **pre-existing** tests needed small adjustments for the now-stricter meta contract, documented inline:
  - The stored-XSS renderer test now **raw-writes** its payload to the DB (bypassing the sanitizer) so it still exercises the renderer's output-escaping — that second line of defense is preserved.
  - Two placeholder Content IDs containing underscores (`'beyondwords_content_id'`, `'test_value'`) were changed to charset-valid values.

## Testing

- `phpcs` (full `WordPress-VIP-Go` scope): **clean** (exit 0).
- **Full PHPUnit suite: `OK — 662 tests, 2079 assertions`**, 0 failures. Ran the whole suite (not just changed files) because the `register_meta` callback registers globally.
- New tests: `sanitize_content_id` unit provider; `ContentId::save` injection cases; a `Sync` test proving `update_post_meta` blanks a crafted ID end-to-end; a `Client` test asserting the URL is `%2F`/`%3F`-encoded.
- Per project guidance, the full Cypress suite was **not** run.

### DIFF
--- a/src/post/class-meta.php
+++ b/src/post/class-meta.php
@@ -180,6 +180,34 @@ class Meta {
 	}
 
 	/**
+	 * Sanitize a BeyondWords Content ID.
+	 *
+	 * Content IDs are BeyondWords-issued numeric IDs or UUIDs, so the only valid
+	 * characters are alphanumerics and hyphens — the same `[a-zA-Z0-9\-]+` charset
+	 * the plugin's own inspect REST route enforces. Anything outside that set is
+	 * rejected (blanked) rather than stored, because the Content ID is later
+	 * interpolated into BeyondWords API URL paths in \BeyondWords\Api\Client, where
+	 * characters such as `/`, `.`, `?`, `#` and `&` would let a crafted value steer
+	 * an authenticated, organization-API-key-scoped request to an attacker-chosen
+	 * endpoint (path/query injection).
+	 *
+	 * Used both directly by the classic-editor save handler and as the
+	 * `sanitize_callback` for the `beyondwords_content_id` post meta, so the
+	 * block-editor / REST write path is validated identically.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @param mixed $content_id Raw submitted Content ID.
+	 *
+	 * @return string Sanitized Content ID, or '' when the value is invalid.
+	 */
+	public static function sanitize_content_id( $content_id ): string {
+		$content_id = sanitize_text_field( (string) $content_id );
+
+		return preg_match( '/^[a-zA-Z0-9-]*$/', $content_id ) ? $content_id : '';
+	}
+
+	/**
 	 * Get the (legacy) Podcast ID for a WordPress Post.
 	 *
 	 * Over time there have been various approaches to storing the Podcast ID.
--- a/src/api/class-client.php
+++ b/src/api/class-client.php
@@ -125,7 +125,7 @@ class Client {
 			return false;
 		}
 
-		$url = sprintf( '%s/projects/%d/content/%s', \BeyondWords\Core\Urls::get_api_url(), $project_id, $content_id );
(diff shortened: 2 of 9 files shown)


src/post/class-meta.php
```diff
@@ -179,6 +179,34 @@ public static function get_content_id( int $post_id, bool $fallback = false ): s
...
+	 * @since 7.0.0
+	 *
+	 * @param mixed $content_id Raw submitted Content ID.
+	 *
+	 * @return string Sanitized Content ID, or '' when the value is invalid.
+	 */
+	public static function sanitize_content_id( $content_id ): string {
+		$content_id = sanitize_text_field( (string) $content_id );
+
+		return preg_match( '/^[a-zA-Z0-9-]*$/', $content_id ) ? $content_id : '';
```
galbus, line 207: @gouravkhunger is this regex OK for content IDs?

gouravkhunger: @galbus Yes, fine. Backend content IDs are only ever a numeric podcast id or a hex UUID (`Podcast::UUID_REGEXP = /\h{8}-(\h{4}-){3}\h{12}/`). `^[a-zA-Z0-9-]*$` is a superset of both, so no valid ID is rejected, while the injection chars (`/ . ? # &`) are still blocked.

gouravkhunger (approved):

PR #582, opened 2026-07-12
fix(site-health): live bounded API probe; collapse timeout consts to two

galbus:
## What

Two related changes to how the plugin bounds blocking requests:

### 1. Site Health probe: bounded and gated, but **live**

Site Health > Info (`SiteHealth::add_rest_api_connection()`) issued a blocking `wp_remote_request()` to the BeyondWords API on every render — with no `timeout` override (WordPress' 5s default), even on fresh installs with no credentials.

- **Skip entirely when unconfigured** — `Utils::has_api_creds()` gate; fresh installs make zero HTTP requests and show *"Not checked — BeyondWords API credentials are not configured."*
- **Bound the request** to the shared 3s ceiling (was an unbounded 5s default).
- **Fixed a latent DNS bug** — the error path passed the full URL to `gethostbyname()` (a no-op). `resolve_api_host()` parses the host first so the *"Unable to reach … at {IP}"* diagnostic reports a real IP. Core passes a bare hostname here too.

**Deliberately NOT cached.** An earlier revision cached the result in a transient; that was reverted after checking prior art: WordPress core runs this same probe **uncached on every Info render** (`WP_Debug_Data::get_wp_core()` → `wp_remote_get('https://wordpress.org', ['timeout' => 10])`). Site Health is a diagnostic — a cached result keeps reporting "unreachable" to an admin who just fixed creds/DNS and hit refresh. No phpcs rule requires caching (`RemoteRequestTimeout` only errors above 3s; `LowExpiryCacheTime` only inspects `wp_cache_*`). `vip_safe_wp_remote_get()` is also avoided: its circuit breaker reports failure without retrying, which would make the diagnostic lie. Same for `Client::call_api()`, which clears `beyondwords_valid_api_connection` on 401 as a side effect.

### 2. Timeout constants: six names for three values → two constants

Per current guidance, **the only genuinely slow BeyondWords endpoint is the voices GET** — content create/update returns the content ID immediately and generates audio asynchronously server-side. So the generous 30s default had no remaining justification, and the pass-through aliases added churn:

| Before | After |
|---|---|
| `Client::REQUEST_TIMEOUT = 30` (default) | removed — default is now 3s |
| `Client::DELETE_TIMEOUT = 3` | removed — uses the default |
| `Client::RENDER_TIMEOUT = 3` | removed — uses the default |
| `Client::VOICES_TIMEOUT = 8` | **`Client::VOICES_REQUEST_TIMEOUT = 8`** |
| `Utils::CONNECTION_CHECK_TIMEOUT = 3` | removed — uses the default |
| — | **`Client::DEFAULT_REQUEST_TIMEOUT = 3`** (default of `call_api()`/`build_args()`/`cached_get()`) |

**Behaviour change (intentional):** content create/read/update, the Magic Embed player bootstrap, and batch delete drop from an implicit 30s timeout to 3s.

## Testing

- `npm run phpcs` (WordPress-VIP-Go): **clean**, no `phpcs:ignore` added.
- PHPUnit `ClientTest` + `SyncTest` + `UtilsTest` + `SiteHealthTest`: **167 tests / 456 assertions passing.** New/updated cases assert: the probe is *not* cached (two renders ⇒ two requests), zero requests without credentials, the probe and all requests use `DEFAULT_REQUEST_TIMEOUT` (≤ 3s per VIP), and voices keeps its longer bound. The old "write paths keep the longer timeout" test premise is replaced by `all_requests_use_the_short_default_timeout`.
- Full Cypress suite not run. The site-health spec's asserted string *"BeyondWords API is reachable"* is unchanged.

### DIFF
--- a/src/api/class-client.php
+++ b/src/api/class-client.php
@@ -60,68 +60,30 @@ class Client {
 	/**
 	 * Default timeout, in seconds, for a BeyondWords API request.
 	 *
-	 * The create/update content calls can be slow, so this is deliberately
-	 * generous. On WordPress VIP those calls are deferred to background cron
-	 * (see `\BeyondWords\Post\Sync::is_async_generation_enabled()`), so the long
-	 * timeout never blocks a web request there.
-	 *
-	 * @since 7.0.0
-	 */
-	const REQUEST_TIMEOUT = 30;
-
-	/**
-	 * Timeout, in seconds, for any request made on a synchronous render path.
-	 *
 	 * VIP's approved ceiling for a blocking remote request — the
 	 * `WordPressVIPMinimum.Performance.RemoteRequestTimeout` sniff errors above
-	 * 3. Every call site that blocks an admin page render shares this single
-	 * value (the constants below, and `Settings\Utils::CONNECTION_CHECK_TIMEOUT`,
-	 * alias it while keeping their own rationale) so the policy can be retuned
-	 * in one place.
+	 * 3 seconds. The API answers fast on every endpoint except voices: content
+	 * create/update returns the content ID immediately and generates the audio
+	 * asynchronously server-side, so even the write paths have no reason to
+	 * wait longer than this.
 	 *
 	 * @since 7.0.0
 	 */
-	const BLOCKING_TIMEOUT = 3;
+	const DEFAULT_REQUEST_TIMEOUT = 3;
 
 	/**
-	 * Timeout, in seconds, for a DELETE request.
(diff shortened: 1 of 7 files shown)


galbus:
@gouravkhunger there are multiple open and recently closed PRs without request timeout values.

This one includes the update to only use two values:

- 3sec default for all requests, as per VIP guidelines
- 8sec or voices query - the only override

If approved, I'll merge the changes here into other open PRs before we review those.

gouravkhunger (approved):

PR #583, opened 2026-07-12
fix(security): stop register_meta leaking secret post meta over REST

galbus:
## Problem

`Sync::register_meta()` registered **every** BeyondWords post-meta key — via `Utils::get_post_meta_keys('all')` (current **and** deprecated) — with `'show_in_rest' => true`.

WordPress returns `show_in_rest` meta in the **`view`** context with **no capability check** (`WP_REST_Meta_Fields::get_value()` performs no auth; `auth_callback` only gates *writes*). So an unauthenticated `GET /wp/v2/posts/<id>` (or the posts collection) returned, in the `meta` object of every compatible published post:

- `speechkit_access_key` — a **legacy per-post API credential** from the v2.x SpeechKit era (on sites upgraded from SpeechKit)
- `beyondwords_error_message` / `speechkit_error_message` — internal BeyondWords API error strings
- `beyondwords_preview_token` — the audio preview token
- `_speechkit_link` / `_speechkit_text` — legacy player URLs and article text
- `speechkit_status` and other internal rows

This is a public information-disclosure bug present on every install.

## Fix

Two layers, both in [`src/post/class-sync.php`](../blob/claude/nifty-dirac-510e9f/src/post/class-sync.php):

**1. Register `show_in_rest` per key, not blanket-true.**
Only the keys the block editor actually reads/writes over REST get `show_in_rest => true`: the 16 `current` keys, plus six deprecated keys the editor still reads for back-compat on SpeechKit-upgraded posts (`REST_LEGACY_META_KEYS` — error/pending notices, play/preview, the Generate-audio toggle and the legacy player link all depend on them). **Every other deprecated key — including `speechkit_access_key`, `_speechkit_text`, cached API responses — is now `show_in_rest => false` and never reaches REST at all.** All keys stay *registered*, so write-sanitisation and Custom-Fields-panel hiding (`is_protected_meta()`) are unchanged.

**2. Strip the still-needed sensitive keys from public responses.**
The keys the editor genuinely needs in the authenticated `edit` context but which must not leak publicly (`REST_PRIVATE_META_KEYS`: `beyondwords_error_message`, `beyondwords_preview_token`, `speechkit_error_message`, `_speechkit_link`) are removed from every non-`edit` response by a new `rest_prepare_{post_type}` filter (`hide_private_meta_from_rest()`). The `edit` context is itself permission-gated by the posts controller, so editors keep full access; anonymous visitors get nothing.

### Why not the simpler "current → true, deprecated → false" split?

Two reasons it wouldn't work:
- `beyondwords_error_message` and `beyondwords_preview_token` are **current** keys but still sensitive — a pure current/deprecated split leaves them exposed.
- The block editor genuinely reads six **deprecated** keys in the `edit` context (e.g. [`error-notice/check.js`](../blob/claude/nifty-dirac-510e9f/src/editor/components/error-notice/check.js), [`open-sidebar/index.js`](../blob/claude/nifty-dirac-510e9f/src/editor/components/open-sidebar/index.js), `play-audio`, `pending-notice`), so blanket-hiding all deprecated keys would regress editing of legacy posts.

The strip layer resolves both: sensitive keys stay available to authenticated editors, hidden from the public.

## Verification

- **`npm run phpcs`** (WordPress-VIP-Go): **PASS**, no `phpcs:ignore` added.
- **PHPUnit `SyncTest`**: **PASS** — `register_meta`, `init`, and a new `register_meta_does_not_expose_private_meta_to_the_public` (3 tests, 54 assertions) run in the real WP test harness.
- Independently confirmed against a live WP REST server: all 39 meta keys land in the correct bucket, all secret/error/token keys are hidden from anonymous `view`, non-sensitive keys stay publicly readable, and an editor still sees the private keys via `context=edit`.
- No JS changed → no Jest/Cypress impact (full Cypress suite intentionally not run).

## Reviewer notes

- Non-sensitive identifiers (`beyondwords_content_id`, `beyondwords_project_id`, and the four non-secret legacy IDs/flags) remain publicly readable — unchanged in sensitivity from before and useful for headless integrations. Only the flagged secret/internal keys are removed or stripped.
- `speechkit_access_key` is now completely absent from REST (both `view` and `edit`) — editors never needed it there; the admin Inspect panel reads it server-side via `has_meta()`.

### DIFF
--- a/src/post/class-sync.php
+++ b/src/post/class-sync.php
@@ -36,10 +36,64 @@ class Sync {
 	const GENERATE_AUDIO_CRON_HOOK = 'beyondwords_generate_audio';
 
 	/**
+	 * Deprecated post-meta keys that still need REST exposure.
+	 *
+	 * On a site upgraded from the legacy SpeechKit plugin these keys can hold a
+	 * post's only BeyondWords data until it is regenerated under v7, so the block
+	 * editor reads them in the authenticated `edit` context as a fallback — the
+	 * error notice, pending notice, play/preview controls, the "Generate audio"
+	 * toggle and the legacy player link all depend on them (see the components
+	 * under [src/editor/components/](src/editor/components/)). They are therefore
+	 * registered with `show_in_rest => true`; every *other* deprecated key —
+	 * including the secret `speechkit_access_key` — is registered
+	 * `show_in_rest => false` and never reaches the REST API.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @var string[]
+	 */
+	const REST_LEGACY_META_KEYS = [
+		'beyondwords_podcast_id',
+		'speechkit_generate_audio',
+		'speechkit_project_id',
+		'speechkit_podcast_id',
+		'speechkit_error_message',
+		'_speechkit_link',
+	];
+
+	/**
+	 * REST-exposed post-meta keys that must never be readable by unauthenticated
+	 * visitors.
+	 *
+	 * The block editor needs these in the authenticated `edit` context, but
+	 * `WP_REST_Meta_Fields` returns `show_in_rest` meta in the public `view`
+	 * context too, with no capability check — so an anonymous
+	 * `GET /wp/v2/posts/:id` would otherwise disclose them. They hold secrets or
+	 * internal data (BeyondWords API error strings, the audio preview token and
+	 * the legacy player URL), so `hide_private_meta_from_rest()` strips them from
+	 * every non-`edit` response.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @var string[]
+	 */
+	const REST_PRIVATE_META_KEYS = [
+		'beyondwords_error_message',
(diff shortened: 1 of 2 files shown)


gouravkhunger (approved):

gouravkhunger:
@galbus there's a merge conflict on this one.

galbus:
@gouravkhunger Conflict resolved — merged latest `main` in (`b1befa6`): main's `DELETE_AUDIO_CRON_HOOK` landed in the same spot in `class-sync.php` as this PR's new REST constants, so I kept both sides (cron hook next to its `GENERATE_AUDIO_CRON_HOOK` sibling, REST constants after); `init()` merged cleanly and registers both actions. Re-verified after the merge: phpcs clean and SyncTest green (80 tests, 185 assertions). GitHub now reports the PR as mergeable.

gouravkhunger (approved):

PR #589, opened 2026-07-19
docs: reduce inline comments to why-only; move rationale into doc/

galbus:
## What

Drastically reduces the volume of inline comments and docblock prose across the plugin — `src/`, root PHP files, `scripts/`, `tests/phpunit/` and `tests/cypress/` — per the "code explains what, comments explain why" rule:

- **PHPDoc / JSDoc**: 1-line summary; a secondary description (max 2 lines) only where behaviour is genuinely non-obvious. All `@since` / `@param` / `@return` / `@var` tags preserved with their existing values.
- **Inline comments**: what-comments removed; why-comments kept, trimmed to 1–2 lines.
- **Functional comments untouched**: `phpcs:ignore/disable/enable`, `/* translators: */`, eslint globals/disables, WordPress dependency section headers, the plugin header in `speechkit.php`, PHPUnit `@test`/`@covers`/`@group`/`@dataProvider` annotations, and Cypress `@group`/`@covers` spec-discovery headers.
- **WordPress hook docblocks** (above `apply_filters`/`do_action`) kept as public API docs, descriptions trimmed to 1 line.

Long-form rationale worth keeping has moved to `doc/` and is linked from the README documentation index:

- `doc/rest-meta-visibility.md` — how BeyondWords post meta is exposed/hidden over REST
- `doc/video-settings-payload.md` — why a full `video_settings` object is sent to the content endpoint
- `doc/preselect-generate-audio.md` — stored format, legacy shapes, no-dirty editor derivation, save-time decision
- `doc/settings-internals.md` — settings-error transport, API connection check contract
- `doc/wordpress-vip.md` — new object-cache notes section
- `doc/running-tests.md` — new PHPUnit/flaky-test lore sections
- README now also links the previously unlisted `async-rest-migration.md` and `legacy-meta-migration.md`

## No lines of code changed

**Every one of the 139 changed PHP/JS files is code-identical to `main`**: byte-identical `php -w` output (comments + whitespace stripped) for PHP, identical espree token streams (JSX-aware, comments excluded) for JS. Formatting around removals was also checked — no introduced trailing whitespace, double blank lines, or blank lines adjacent to braces (pattern counts compared per-file against `main`).

`phpcs` (WordPress-VIP-Go ruleset) and `wp-scripts lint-js` both pass.

## Includes latest `main`

[...]

PR #590, opened 2026-07-19
refactor: cut the inline commentary added with the polling change

galbus:
Follow-up to #566 (already merged). Comment-only cleanup — **no behaviour change**.

Comments should say *why*, not *what*. #566 landed with a lot of narration and JSDoc prose that just restated the code; this drops it and keeps only the non-obvious rationale, at one or two lines each.

## What went

- **`src/editor/lib/poll-content-status.js`** — the worst offender (91 comment lines of 179). Dropped the per-constant docblocks, the "what it does" prose and the loop narration → 36 of 121.
- **`play-audio/hooks.js`** — 16-line hook docblock → a 2-line summary; the 5-line `mountContentIdRef` explainer → 2; deleted the block comments restating the `if`/`else`.
- **`content-id/classic-metabox.js`** — each helper's docblock trimmed to a one-liner + `@param`s; dropped the "skip while hidden" and "SDK constructor can throw" comments (the code already says both).
- **`classic/class-metabox.php`** — the two 14-line block comments → 2 lines each.
- **`tests/phpunit/.../test-metabox.php`** — both test docblocks → 3 lines; dropped the assertion narration and trimmed the `store_raw_content_id` note.

## What stayed (the non-obvious *why*)

- 404-caching is why we poll before embedding.
- Hidden cycles never fetch, so they must not spend the budget.
- One blip shouldn't abandon the wait.
- Unknown statuses count as terminal so the loop always ends.
- The SDK is only emitted for posts that had content at load (why we clear the spinner).
- The HEX flags stop an untrusted value closing the JS string literal.
- Why the block poll isn't Cypress-covered (SDK load isn't deterministic in CI).

## Deliberately untouched

The 5-line namespace comment in `hooks.js` and the best-effort-save comment in `classic-metabox.js` are both pre-existing in `main` — trimming them would drag unrelated code into this diff.

## Verification

`lint:js`, `webpack build` and `php -l` all pass. Diff is **-209/+44** and verifiably comment-only (no non-comment line changed).

### DIFF
--- a/src/editor/classic/class-metabox.php
+++ b/src/editor/classic/class-metabox.php
@@ -221,20 +221,8 @@ class Metabox {
 
 		// phpcs:disable WordPress.WP.EnqueuedResources.NonEnqueuedScript
 		if ( ! empty( $content_id ) ) :
-			/*
-			 * REST-API content may still be processing on the BeyondWords backend.
-			 * Embedding the player now would 404 — and the CDN would cache that
-			 * 404 — for content that isn't ready, so render a loading state and
-			 * let classic-metabox.js poll GET /content until the status is
-			 * `processed`, then construct the player.
-			 *
-			 * The Content ID is editable post meta and the preview token comes
-			 * from the REST API, so both are untrusted here. They are emitted as
-			 * esc_attr()'d data-* attributes and read back with getAttribute(),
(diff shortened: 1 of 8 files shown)


galbus:
This is now in https://github.com/beyondwords-io/wordpress-plugin/pull/589

PR #591, opened 2026-07-19
docs: refresh stale content in doc/ against the v7 codebase

galbus:
## Why

Audited every file in `doc/` against the current v7 code. Several docs predate the v7 rewrite and now describe things that no longer exist. Each finding below was verified against the source before rewriting — no doc text was changed on a hunch.

## Substantially stale

**`wp-config.md`** — documented `BEYONDWORDS_AUTO_SYNC_SETTINGS`, a constant that **no longer exists anywhere in the plugin**. `grep -rn BEYONDWORDS_AUTO_SYNC_SETTINGS --include='*.php' src/ *.php` returns nothing; the only occurrence in the repo is a leftover env var in `.wp-env.tests.json` that nothing reads. The settings-sync feature it controlled was removed in v7. Anyone defining it in `wp-config.php` today gets silence.

Removed that section, and documented the **six URL-override constants that do work** (`Core\Urls`), with their real defaults and the `strlen()` guard that makes an empty-string override a no-op. Also tightened `BEYONDWORDS_AUTOREGENERATE` — the check sits inside the "already has a content ID" branch, so it suppresses *re*generation only, not initial generation.

**`plugin-features.md`** — written 2.5 years ago, predates v7 entirely. Listed the "Display player" control that v7 replaced with the Embed dropdown, and omitted most of the current surface (Integration method, Player UI, video output, voice selection, the Delete audio bulk action, per-block generation toggles, Site Health).

## Corrections

| Doc | Was wrong |
|---|---|
| `async-rest-migration.md` | Said errors are not cached — they are negative-cached for `CACHE_TTL_ON_ERROR`. Named `Client::get_summarization_settings()`, which doesn't exist. Listed background delete as "deferred" after it shipped. Still written in unmerged-PR voice. |
| `wordpress-vip.md` | Described the uninstaller's transient cleanup as deleting known keys; it's a `LIKE '_transient_beyondwords_%'` SQL prefix sweep. The VIP-relevant point is that object-cache hosts have no such rows, so the sweep matches nothing and entries expire by TTL. |
| `running-tests.md` | Told developers to create `.wp-env.override.json` for the **tests** env; the correct file is `.wp-env.tests.override.json`. No Jest section despite Jest tests and a CI Jest job. Presented the full 20-minute Cypress suite as the normal workflow instead of the per-spec `@covers`/`@group` flow in AGENTS.md. |
| `legacy-meta-migration.md` | Headline downgrade-safety guarantee didn't note that the migration **deletes** `beyondwords_disabled`. Described `play-audio/` and `select-voice/` as removed/replaced — both still exist and are wired up. |
| `settings-internals.md` | Omitted the 5-minute `CONNECTION_CHECK_TTL` throttle (keyed on a credential fingerprint), which is what actually governs when validation re-runs. |
| `getting-started.md` | Step 5 (`npm run composer -- install`) ran before step 6 (`npm run env:start`), but the composer script dispatches into a wp-env container that doesn't exist yet — unrunnable on a fresh clone. |
| `code-quality-checks.md` | Listed `test_phpunit`/`coverage_check` as commit-time checks when they aren't in that testsuite; omitted `lint_plugins`; overstated what CI runs under GrumPHP. |
| `preselect-generate-audio.md` | Mode is `'off'`, not `'none'`. Didn't mention `DEFAULT_VALUE` makes a fresh install preselect for posts. Misdescribed what the classic metabox JS does. |
| `rest-meta-visibility.md` | Over-broad "voice config is never exposed"; omitted `register_rest_meta_visibility()`; mislabelled a BeyondWords-era key as a SpeechKit one. |
| `video-settings-payload.md` | Only covered the branch where a Video size is chosen; didn't document the project-default fallback. Framed an unreleased dev state as "earlier plugin versions". |

## Notes for review

- Docs-only — no source changes.
- Two of these (`wordpress-vip.md`, `settings-internals.md`) were errors I introduced in #589 when extracting rationale out of code comments; they're fixed here.
- Every relative link in the changed docs resolves, and code fences are balanced (checked).


gouravkhunger (approved):
Thanks @galbus! There are some conflicts needing to be fixed before this can be merged.

galbus:
@gouravkhunger there's been quite a few PRs merged in since this was open so I'll resolve conflicts, then give it 1 more pass before merging

PR #599, opened 2026-07-26
fix: fall back to the default asset when a stored Embed is invalid

galbus:
## Problem

When a stored `beyondwords_embed` became invalid after a Source or Output change, all three layers resolved it to **None**, so the post published with no player and no warning:

- classic: `recomputeEmbed()` in `src/editor/components/settings-fields/classic-metabox.js`
- block: the `useEffect` in `src/editor/components/settings-panel/player-section.js`
- PHP: `get_effective_embed()` in `src/editor/components/settings-fields/class-settings-fields.php`

The rule was consistent, but the editors exposed it differently. Block keeps an untouched Embed unset in meta and live-derives the default, so the rule rarely fired. Classic rendered the derived default into the form and persisted it on publish, so a user who never touched Embed still hit the None fallback after changing Output — same action, different outcome per editor.

Reproduction (classic): new post → Embed shows Audio (post) → change Output to Video → Embed rebuilds and selects None → publish → no player.

## What changed

**1. Invalidation now resolves to the first produced asset, not None.**

| Layer | Location |
|---|---|
| PHP | `get_effective_embed()` — `class-settings-fields.php` |
| Block | the `useEffect` — `player-section.js` |
| Classic | `recomputeEmbed()` — `classic-metabox.js` |

An explicit **None** is unaffected: None is a valid option for every Source × Output, so it never reaches the invalidation branch. That is precisely what makes it the deliberate opt-out, and it is now covered by a test in each layer.

**2. Classic no longer persists an Embed the user did not choose.**

The Player section renders a `beyondwords_embed_touched` hidden input, `classic-metabox.js` sets it on a user `change` of the Embed select, and `save()` drops `beyondwords_embed` from its key list when the flag is empty.

A Source/Output rebuild deliberately does **not** set the flag — a rebuild is not a choice — so an untouched Embed stays unset in meta and is re-derived at render, matching the block editor. This also resolves the wider asymmetry where block posts followed later default changes while classic posts were pinned to publish-time values.

## Reviewer notes

- **Classic can no longer set Embed with JS disabled**, since the touched flag is set by JS. `classic-metabox.js` is already required for the Script template / Video field show-hide behaviour, so classic is already degraded without it — but this widens that. It is inherent to the unified-persistence approach, not an oversight.
- **The flag is wired before `init()`'s early return**, because it gates whether `save()` persists the Embed at all; it must be set even if Source/Output are somehow absent.
- **Two existing tests encoded the old rule and are corrected.** `merge_post_settings_embed_invalid_for_source_output_falls_back_to_none` asserted only *key absence*, so it would have passed either way — it moves to Script × Video, where the default asset sets both `video` and `summary`, so a regression now actually fails. A companion test proves an explicit None survives the same invalidation.
- The front end was never at risk regardless, because `get_effective_embed()` re-validates at render time.

## Testing

- **PHPUnit** — 720 tests pass, 1 skip (the known multisite-only case). Adds direct `get_effective_embed()` cases (including the previously untested unset + legacy `beyondwords_disabled` branch), `save()` cases for both sides of the flag, and a check that the hidden input renders.
- **Jest** — 50 tests pass, 8 new in `classic-metabox.test.js`. The classic module is a bare IIFE, so the suite boots it against jsdom and drives it with real `change` events rather than adding test hooks to production code.
- **Cypress** — `classic-editor/settings-fields` 24/24, `block-editor/settings-panel` 18/18, both `display-player` specs 6/6. New cases cover invalidation landing on the default asset, an explicit None surviving, and (classic) an untouched Embed staying out of meta while a chosen one persists, asserted via `getPostMeta`.
- **Lint** — `phpcs` clean on the changed PHP, `lint:js` clean across `./src`.

The invalidation fallback previously had no test in the block or classic layer at all; it does now in all three.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -128,6 +128,8 @@ Release date: tbc
 * [#542](https://github.com/beyondwords-io/wordpress-plugin/pull/542) Surface a `WP_Error` from `get_content()` instead of a fatal `TypeError`.
 * [#588](https://github.com/beyondwords-io/wordpress-plugin/pull/588) Ship `symfony/dom-crawler` 5.4.52 to fix CVE-2026-45071 (XXE / local file disclosure).
     * The composer constraint now floors at the patched release, so a vulnerable version can no longer be bundled.
+* [#599](https://github.com/beyondwords-io/wordpress-plugin/pull/599) Keep the player when a Source or Output change invalidates the stored "Embed".
+    * In the classic editor an "Embed" you never chose is no longer saved with the post, so those posts now follow later Source and Output changes.
 
 **Deprecations**
 
--- a/src/editor/components/settings-fields/class-settings-fields.php
+++ b/src/editor/components/settings-fields/class-settings-fields.php
@@ -280,10 +280,8 @@ class SettingsFields {
 	/**
 	 * Resolve the effective Embed value for a post.
 	 *
-	 * Centralised so the shown default and the rendered player (ConfigBuilder)
-	 * never diverge. Legacy opt-outs resolve to None; a value that no longer fits
-	 * the current Source × Output falls back to the default asset, because only an
-	 * explicit "None" — always a valid option — means "show no player".
+	 * Centralised so the shown default and the rendered player (ConfigBuilder) never
+	 * diverge; legacy opt-outs resolve to None, stale values to the default asset.
 	 *
 	 * @since 7.0.0
 	 *
@@ -419,9 +417,6 @@ class SettingsFields {
 	/**
 	 * Render the Player section fields: Embed.
 	 *
-	 * The select shows the effective value, which for an unset Embed is the derived
-	 * default — so the accompanying flag records whether the user actually chose one.
-	 *
 	 * @since 7.0.0
 	 *
 	 * @param \WP_Post $post The post object.
@@ -442,6 +437,7 @@ class SettingsFields {
 				'speechkit'
 			)
 		);
+		// Lets save() tell a real choice from the rendered default.
 		?>
 		<input type="hidden" id="beyondwords_embed_touched" name="beyondwords_embed_touched" value="" />
 		<?php
@@ -607,10 +603,7 @@ class SettingsFields {
 			'beyondwords_embed',
 		];
 
-		// The Embed is only persisted once the user picks one, because the select
-		// always submits a value — the derived default when untouched. Leaving it
-		// unset keeps classic posts in step with block posts and lets a later
-		// Source/Output change re-derive the default asset.
+		// Only a real choice is stored; an unset Embed is re-derived at render.
 		if ( empty( $_POST['beyondwords_embed_touched'] ) ) {
 			$keys = array_values( array_diff( $keys, [ 'beyondwords_embed' ] ) );
 		}
--- a/src/editor/components/settings-fields/class-settings-fields.php
+++ b/src/editor/components/settings-fields/class-settings-fields.php
@@ -281,7 +281,9 @@ class SettingsFields {
 	 * Resolve the effective Embed value for a post.
 	 *
 	 * Centralised so the shown default and the rendered player (ConfigBuilder)
-	 * never diverge; legacy opt-outs and no-longer-valid values resolve to None.
+	 * never diverge. Legacy opt-outs resolve to None; a value that no longer fits
+	 * the current Source × Output falls back to the default asset, because only an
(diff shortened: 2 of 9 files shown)


gouravkhunger:

Logic is correct across all three layers (None always valid so the explicit opt-out survives; `default_embed()` always yields an asset), security is clean (touched flag read after nonce + `edit_post`), and coverage is strong — but the **changelog entry for #599 is missing** from `readme.txt`, which is a project non-negotiable and must be added under `**Fixes**` before merge.

<details>

## Architecture map

Bug: a stale `beyondwords_embed` (invalid after a Source/Output change) resolved to **None** in all three layers, so a post published with no player and no warning. Classic hit it easily because it rendered the derived default into the form and persisted it on publish.

Fix, two parts:
1. Invalidation now falls back to the **first produced asset**, not None — PHP `get_effective_embed()`, block `useEffect`, classic `recomputeEmbed()`.
2. Classic stops persisting an Embed the user didn't pick — new `beyondwords_embed_touched` hidden input; `save()` drops `beyondwords_embed` when the flag is empty. Matches the block editor's unset-and-re-derive model.

## Affected subsystems

- `src/editor/components/settings-fields/` — PHP (`class-settings-fields.php`), classic JS (`classic-metabox.js`), new Jest suite.
- `src/editor/components/settings-panel/player-section.js` — block editor effect.
- Tests: PHPUnit, Jest, Cypress (block + classic).

## Dependency graph

No new deps. `player-section.js` drops the now-unused `EMBED_NONE` import and reuses the existing `getDefaultEmbed(source, output)` from `helpers.js` — signature matches. Classic adds a local `getDefaultEmbed(options)` mirroring helpers (takes pre-derived options). No cross-module coupling added.

## Risky files

`class-settings-fields.php` `get_effective_embed()` + `save()` — the behavior change touching both render and persist. Reviewed and sound: `None` is always the first entry in `embed_options()`, so `is_embed_valid(None)` is always true → an explicit opt-out is never rebuilt. `default_embed()` always returns an asset (every valid Source×Output yields ≥1 non-None option).

## Public API / DB / auth

- **DB:** no schema change. Semantic change — classic now leaves `beyondwords_embed` meta unset when untouched (was persisting the derived default). Front end unaffected: `get_effective_embed()` re-derives at render.
- **Auth/security:** the touched-flag read sits after the nonce check and `current_user_can('edit_post')` in `save()`. No new trusted input. Fine.
- **Public API:** none.

## Behavior change worth noting

Posts that previously published with no player (silent None fallback) will now show a player on the default asset — the intended fix. Explicit None is preserved, so no user opt-out is overridden. No infinite-loop risk in the effect (deps `[source, output]` only).

## Reviewer-note concern (minor, acknowledged)

Classic can no longer set Embed with **JS disabled** — the flag is JS-set, so `save()` never persists an Embed choice without JS. The PR flags this; `classic-metabox.js` is already required for show/hide behavior, so it widens an existing degradation rather than introducing a regression from a working state. Acceptable.

## Blocker — missing changelog entry

No entry for **#599** in `readme.txt` under the `= 7.0.0 =` unreleased block. AGENTS.md non-negotiable: every PR is listed in the changelog before it merges. This is a user-facing fix → `**Fixes**`, one line, e.g.:

```
* [#599](https://github.com/beyondwords-io/wordpress-plugin/pull/599) Keep the player on the default asset when a stored Embed is invalidated by a Source/Output change.
```

## Tests

Strong. The fallback previously had zero tests in the block/classic layers; now covered in all three. Two old tests that encoded the None rule are corrected (moved to Script×Video so a regression actually fails). Explicit-None survival is tested per layer. Cypress asserts unset meta via `getPostMeta`.

## Review priority order

1. `class-settings-fields.php` `get_effective_embed()` / `save()` — core logic + persist semantics.
2. `player-section.js` + `classic-metabox.js` — parity across editors.
3. Missing changelog entry — add before merge.
4. Test files.

## Verdict

Code is correct, well-scoped, well-tested, and security clean. Single blocker: the missing changelog entry (project non-negotiable). Add it and this is mergeable.

</details>

galbus:
Added the changelog entry in `baa9eee` — one bullet under `= 7.0.0 =` → `**Fixes**`, plus a single sub-bullet for the changed classic-editor persistence (the one thing the title can't carry), per AGENTS.md rule 5. The branch is also rebased onto the merged `main`, which brought that rule in.

gouravkhunger (approved):

PR #601, opened 2026-07-26
Remove the Script-only Source option (v7)

galbus:
## Problem

The Source dropdown offered Post / Script / Post + script in both editors, but **script-only was never a real mode**. The script is derived from the post body, so the platform always generates the post assets — the plugin's own payload code confirmed it: `script` and `post_and_script` built an identical request in `src/post/class-content.php`, and nothing suppressed post generation.

Verified on staging (project 54024, 7.0.0-beta.1): every Source = Script post produced article assets. Post 3468 article audio; post 3446 article audio + article video; post 3442 article audio + script audio once the script generated.

Consequences of keeping the option:

- The UI promised script-only output while the account was billed for the post assets too.
- The Embed dropdown for Source = Script derived only None / Audio (script) / Video (script), hiding the post assets that actually existed on the content item.

## What changed

**Source offers Post / Post + script**, in `getSourceOptions()` ([helpers.js](src/editor/components/settings-panel/helpers.js), block), `source_options()` ([class-settings-fields.php](src/editor/components/settings-fields/class-settings-fields.php), classic), and the constants in [classic-metabox.js](src/editor/components/settings-fields/classic-metabox.js). `SOURCE_SCRIPT` became `LEGACY_SOURCE_SCRIPT` in both languages — the literal is still needed to normalise.

**Normalisation on read.** New `normalizeSource()` (JS) and `normalize_source()` + `get_source( $post_id )` (PHP) resolve `script`, unknown values, and unset to a source the dropdown offers. Wired into both block sections, both classic renderers, `get_effective_embed()`, and the payload builder — so a stale `script` in meta can never render a blank select.

**Payload.** `class-content.php` dropped its hard-coded `[ 'script', 'post_and_script' ]` for `SettingsFields::source_includes_script()`, so this logic no longer has a raw-literal fourth copy.

**`sourceIncludesPost()` deleted** (JS, PHP, classic JS). With script-only gone, every remaining source generates the post, so it was a tautology; `getEmbedOptions()` / `embed_options()` now push the post assets unconditionally — which is the reported bug, stated as code.

**Migration.** `migrate_source_script_to_post_and_script()` in the `< 7.0.0` block rewrites the meta, batched and idempotent like its sibling.

## Reviewer notes

**One judgement call worth a look:** the migration also pins `beyondwords_embed` to the script asset when the post had no explicit choice. Script-only derived script embeds alone, so the effective default was `audio_script`; under Post + script the default becomes `audio_post`. Without the pin, a beta post would silently switch which asset its player shows on upgrade.

**No "section 2.2 Source" exists in this repo** — no doc contains the Source option list or the Embed examples (closest is `readme.txt:92`, field names only). That amendment belongs wherever the project overview actually lives. The removal and migration are documented in [doc/legacy-meta-migration.md](doc/legacy-meta-migration.md) instead.

`languages/speechkit.pot` still carries the `"Script"` msgid; CI regenerates it via `composer wp:i18n:make-pot` at release build.

Comments and docblocks follow the § Documentation rules from `claude/agents-txt-docs-guidelines-dc763e` (one line per inline comment, one-line docblock summaries, long-form in `doc/`).

## Testing

- `npm run test:unit` (settings-panel helpers) — 28 pass, including new `normalizeSource` and legacy-value cases
- Full PHPUnit — **717 pass, 1 skipped** (the pre-existing multisite Uninstaller skip). New coverage: `normalize_source`, `get_source`, legacy-source rendering, stale-form rejection on save, and the migration.
- `phpcs` (WordPress-VIP-Go) clean, no new `phpcs:ignore`; `npm run lint:js` clean

**Cypress specs are updated but were not run** — `block-editor/settings-panel.cy.js` and `classic-editor/settings-fields.cy.js` need a dev env serving this branch:

```bash
npm run cypress:run -- --browser chrome --spec 'tests/cypress/e2e/block-editor/settings-panel.cy.js,tests/cypress/e2e/classic-editor/settings-fields.cy.js'
```

### DIFF
--- a/doc/legacy-meta-migration.md
+++ b/doc/legacy-meta-migration.md
@@ -35,6 +35,21 @@ To carry existing opt-outs forward, the v7.0.0 migration in
 flag. `Player::is_enabled()` still honours `beyondwords_disabled` as a fallback
 for any post the migration hasn't reached (e.g. after a downgrade/re-upgrade).
 
+### `beyondwords_source = script` → `post_and_script`
+
+Source offered Post / Script / Post + script during the 7.0.0 betas, but
+script-only was never a real mode: the script is derived from the post body, so
+the platform generates the post assets either way (verified on staging — every
+Source = Script post produced article assets). The Embed dropdown then derived
+script assets only, hiding post assets that existed on the content item.
+
+Script was dropped from the option list. `SettingsFields::normalize_source()`
+and `normalizeSource()` (settings-panel/helpers.js) resolve `script`, unknown,
+and unset values to a source the dropdown offers, and the v7.0.0 migration
+`migrate_source_script_to_post_and_script()` rewrites the meta. Posts with no
+explicit Embed get one pinned to the script asset first, so the upgrade doesn't
+switch which asset the front-end player shows.
+
 ### Components removed
 
 - `src/editor/components/player-style/`
--- a/readme.txt
+++ b/readme.txt
@@ -128,6 +128,9 @@ Release date: tbc
 * [#542](https://github.com/beyondwords-io/wordpress-plugin/pull/542) Surface a `WP_Error` from `get_content()` instead of a fatal `TypeError`.
 * [#588](https://github.com/beyondwords-io/wordpress-plugin/pull/588) Ship `symfony/dom-crawler` 5.4.52 to fix CVE-2026-45071 (XXE / local file disclosure).
     * The composer constraint now floors at the patched release, so a vulnerable version can no longer be bundled.
+* [#601](https://github.com/beyondwords-io/wordpress-plugin/pull/601) Remove the "Script" option from the post Source setting, which offers "Post" or "Post + script".
+    * A script is written from the post, so the post's own audio and video were always generated (and billed) alongside it — while the Embed setting offered only the script assets, hiding the post assets that existed.
+    * Posts already saved as "Script" are migrated to "Post + script" on upgrade, and keep embedding their script asset.
 
 **Deprecations**
 
--- a/readme.txt
+++ b/readme.txt
@@ -128,15 +128,14 @@ Release date: tbc
 * [#542](https://github.com/beyondwords-io/wordpress-plugin/pull/542) Surface a `WP_Error` from `get_content()` instead of a fatal `TypeError`.
 * [#588](https://github.com/beyondwords-io/wordpress-plugin/pull/588) Ship `symfony/dom-crawler` 5.4.52 to fix CVE-2026-45071 (XXE / local file disclosure).
     * The composer constraint now floors at the patched release, so a vulnerable version can no longer be bundled.
-* [#601](https://github.com/beyondwords-io/wordpress-plugin/pull/601) Remove the "Script" option from the post Source setting, which offers "Post" or "Post + script".
-    * A script is written from the post, so the post's own audio and video were always generated (and billed) alongside it — while the Embed setting offered only the script assets, hiding the post assets that existed.
-    * Posts already saved as "Script" are migrated to "Post + script" on upgrade, and keep embedding their script asset.
 
 **Deprecations**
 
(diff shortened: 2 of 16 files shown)


gouravkhunger:

Code is solid — the fake script-only mode is correctly diagnosed and killed, with read-side normalisation, save-side rejection and a batched idempotent migration all tested. **The one required change: there is no changelog entry for this PR in `readme.txt` (AGENTS.md non-negotiable rule 4)** — this is user-visible, so it needs a line in the unreleased 7.0.0 block.

<details>

## What it does
Removes the "Script" source mode. The script is derived from the post body, so the platform always generates the post assets — script-only was never a real mode; the UI promised script-only output while the account was billed for post assets, and the Embed dropdown hid the post assets that actually existed. Source now offers **Post / Post + script**.

## Affected subsystems
- **Settings fields** (PHP + block JS + classic JS) — source/embed option builders
- **Content payload** (`class-content.php`) — summarization-enable logic
- **Updater** — new v7.0.0 data migration
- **Tests** — jest, PHPUnit, Cypress specs
- **Docs** — `doc/legacy-meta-migration.md`

## Dependency graph changes
- `SOURCE_SCRIPT` renamed to `LEGACY_SOURCE_SCRIPT` (PHP const + JS export); the literal is kept only for normalisation/migration.
- `source_includes_post()` / `sourceIncludesPost()` deleted in all three places — now a tautology since every source generates the post.
- New: `normalize_source()` / `normalizeSource()`, `get_source( $post_id )`, `migrate_source_script_to_post_and_script()`.
- `class-content.php` now calls `SettingsFields::source_includes_script()` instead of a raw `['script','post_and_script']` literal — removes the fourth duplicated copy of that logic. Good dedup.

## Public API changes
Internal only. No documented hooks/filters removed. `languages/speechkit.pot` still carries the "Script" msgid — regenerated by CI at build. Low external risk.

## DB changes
Migration rewrites `beyondwords_source` meta `script` → `post_and_script`, batched at 100, idempotent (a migrated post no longer matches the meta_query). It also pins `beyondwords_embed` to the script asset when the post had no explicit choice, so the upgrade doesn't silently switch which asset the player shows. The judgement call is reasonable and clearly flagged.

## Auth/security
No change. Save path keeps nonce + `current_user_can( 'edit_post' )`. A stale form posting `script` is now rejected on save (`is_valid_meta_value` validates against `source_options()`, which dropped script), and the read-side `normalize_source` resolves any already-stored stale value. Defence-in-depth is correct.

## Deployment/config
Migration fires in the `< 7.0.0` updater block. The `meta_query` carries `phpcs:ignore WordPress.DB.SlowDBQuery` — the same pattern as the sibling `migrate_disabled_to_embed_none`, so it's consistent with existing convention, not a new exception.

## Risky files (priority order)
1. `src/core/class-updater.php` — one-time DB write plus the embed-pin judgement. Batch loop termination and pin logic worth a close look; covered by new `test-updater.php`.
2. `src/editor/components/settings-fields/class-settings-fields.php` — `normalize_source`/`get_source` wired into four render paths plus payload and save validation. Core logic.
3. `helpers.js` / `classic-metabox.js` — must stay in lockstep with the PHP; they do (a test asserts the legacy value derives identical options).

## Major flag
- **No changelog entry.** The unreleased 7.0.0 block in `readme.txt` has no `#601` line (grep confirms absent). AGENTS.md non-negotiable rule 4 requires every PR to be listed before merge. This is user-visible (removed setting + billing behaviour) and belongs under **Enhancements** or **Fixes**, likely with a migration sub-bullet.

## Minor
- Cypress specs are updated but were not run (needs a dev env on this branch). Acceptable given the mid-rewrite note, but the e2e coverage is unverified.

## Verdict
Code quality is high. The bug is correctly diagnosed and fixed, with three layers (read normalisation, save rejection, migration) all tested, PHP↔JS parity held, and rationale documented in `doc/` rather than in comments. The one required change before merge is the changelog entry.

</details>

galbus:
Added the changelog entry — filed under **Deprecations** (removed setting) with a single migration sub-bullet, per AGENTS.md Documentation rule 5; also merged `main` in so the branch carries the rule.

gouravkhunger (approved):

PR #603, opened 2026-07-26
fix(classic-editor): show Generating state while generation is queued

galbus:
## Problem

Publishing in the classic editor could leave the Player section **completely empty** until the author reloaded the page — no player, no spinner, no message. The metabox looks like generation failed when it is simply still running, which misleads testing and support.

The polling delivered under [S-8779](https://linear.app/beyondwords/issue/S-8779/poll-rest-api-from-wordpress-plugin-to-ensure-player-is-ready) (#566) already covers part of this, but there are **two** distinct waits after a publish and only the second was handled:

| Wait | Before | After |
|---|---|---|
| WordPress has no content ID yet (job queued) | Player section renders **empty** | "Generating…" + poll |
| Content ID exists, BeyondWords still processing | "Generating…" + poll | unchanged |

`render_meta_box_content()` gated the entire player area on `has_content()`, so during the first wait it emitted the "Player" heading and nothing else. The existing poll could not help: `initMetaboxPlayer()` bails without a `data-content-id`, and that container was never rendered in the first place.

The tell-tale symptom is that reloading shows the **finished player** directly — never the Generating state. If a content ID had existed at first render you would have seen the spinner.

This happens whenever generation is deferred to cron — `Sync::is_async_generation_enabled()`, i.e. WordPress VIP or the `beyondwords_async_generate_audio` filter. It is pinned by the existing `on_add_or_update_post_defers_to_cron_when_async_enabled` test, which asserts `beyondwords_content_id` is `''` immediately after the save.

## Changes

- **`src/editor/classic/class-metabox.php`** — new `awaiting_generation_embed()` renders the loading container with `data-post-id` / `data-await-content` when a generation job is queued but no content ID exists. Gated on `wp_next_scheduled( Sync::GENERATE_AUDIO_CRON_HOOK, … )` so posts with no audio coming stay blank rather than showing a spinner that would never resolve.
- **`src/editor/components/content-id/classic-metabox.js`** — new `awaitContentId()` polls the post's own meta over `wp/v2` until the cron job writes an ID, then sets the data attributes and hands off to `initMetaboxPlayer()`.

`awaitContentId()` reuses the existing `pollContentStatus()` by reporting `'queued'` (already a non-terminal status) until the ID lands, so it inherits the 120s budget, hidden-tab handling and supersede token rather than duplicating them.

The handoff matters: the player still waits for `processed` rather than embedding straight away, so a still-processing asset never 404s into the CDN cache.

No new REST route is needed — `beyondwords_content_id` is already `show_in_rest` and is not in `REST_PRIVATE_META_KEYS`, so the `wp/v2` request the Fetch button already makes can serve it.

## Testing

- **PHPUnit: 714/715 pass.** Full `MetaboxTest` green (15 tests), including two new cases — renders the loading state when a job is queued, and renders nothing when one is not.
- The single failure is `build/index.asset.php: No such file or directory` in `src/editor/block/`, a file this PR does not touch. Confirmed pre-existing by stashing the changes and re-running: it fails identically on a clean tree. It is the gitignored `build/` dir being absent without `npm run build`.
- ESLint and phpcs clean on all changed files.

### Reviewer note

✅ **The two new Cypress specs have now been run in CI and pass** on both PHP 8.0 and 8.5:

```
✓ shows a Generating state while deferred generation is still queued
✓ hands off to the content-status poll once the queued job writes an ID
```

This supersedes the earlier note on this PR, which said they were unrun — at the time they could only be run against an env that mounted `main`, so they would have passed without exercising this branch. CI runs them against the branch properly. To run locally:

```
npm run cypress:run -- --spec 'tests/cypress/e2e/classic-editor/content-id.cy.js'
```

They rely on a new `scheduleAudioGeneration` Cypress task (`cypress.config.js`) that queues the cron event the async path would create, scheduled an hour out so WP-Cron cannot fire it mid-test.

### Scope

This fixes the deferred-generation path. On a non-VIP site generation is a blocking call during the save, so the content ID is already present and the existing polling works. If the empty Player section is ever reproduced on a **non-VIP** site, that is a different fault — most likely the 3s `DEFAULT_REQUEST_TIMEOUT` in `src/api/class-client.php` failing the create call — and should be tracked separately.

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -227,9 +227,7 @@ function setupNodeEvents( on, config ) {
 			return null;
 		},
 
-		// Queue the deferred generation job the async (VIP) path would create.
-		// Scheduled far enough ahead that WP-Cron can't run it mid-test — the
-		// metabox only needs wp_next_scheduled() to be truthy.
+		// An hour out, so WP-Cron can't actually run the job mid-test.
 		async scheduleAudioGeneration( postId ) {
 			await execWp(
 				`eval 'wp_schedule_single_event( time() + 3600, "beyondwords_generate_audio", [ ${ parseInt(
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -227,6 +227,19 @@ function setupNodeEvents( on, config ) {
 			return null;
 		},
 
+		// Queue the deferred generation job the async (VIP) path would create.
+		// Scheduled far enough ahead that WP-Cron can't run it mid-test — the
+		// metabox only needs wp_next_scheduled() to be truthy.
+		async scheduleAudioGeneration( postId ) {
+			await execWp(
+				`eval 'wp_schedule_single_event( time() + 3600, "beyondwords_generate_audio", [ ${ parseInt(
+					postId,
+					10
+				) } ] );'`
+			);
+			return null;
+		},
+
 		async getPostMeta( options ) {
 			const { postId, metaKey } = options;
 			try {
--- a/doc/async-rest-migration.md
+++ b/doc/async-rest-migration.md
@@ -38,6 +38,26 @@ returns immediately.
 - Error meta is written when the job runs (≈1 cron tick later) rather than
   inline, so the editor surfaces an API error on its next load.
 
+### Editor state while the job is queued
+
+Deferral means the post-publish page load has **no content ID at all**, so there
+is nothing for the content-status poll (`#566`) to query. Left alone the classic
+metabox rendered the "Player" heading and nothing else — no player, no spinner,
+no message — until the author reloaded.
+
+`Metabox::awaiting_generation_embed()` covers that window: when
+`wp_next_scheduled( Sync::GENERATE_AUDIO_CRON_HOOK, … )` is set but no content ID
+exists, it renders the loading container with `data-post-id` /
+`data-await-content`. `awaitContentId()` in `content-id/classic-metabox.js` then
+polls the post's own meta over `wp/v2` until the cron job writes an ID, sets the
+data attributes and hands off to `initMetaboxPlayer()` — so the player still
+waits for `processed` rather than embedding a 404 the CDN would cache.
+
+It reuses `pollContentStatus()` by reporting `queued` (already a non-terminal
+status) until the ID lands, inheriting the time budget, hidden-tab handling and
+supersede token. Gating on the scheduled job matters: a post with nothing queued
+has no audio coming, so a spinner there would never resolve.
+
 ## Deferred (follow-up)
 
 ### Classic-editor async reads
--- a/readme.txt
+++ b/readme.txt
@@ -109,6 +109,7 @@ Release date: tbc
 
 **Fixes**
 
+* [#603](https://github.com/beyondwords-io/wordpress-plugin/pull/603) Show a loading state in the classic editor while audio generation is still queued.
 * [#564](https://github.com/beyondwords-io/wordpress-plugin/pull/564) Send the full `video_settings` payload so videos generate.
     * Selecting "Video" or "Audio + video" output now sends the complete video settings (seeded from the project defaults), fixing posts that produced no video.
 * [#564](https://github.com/beyondwords-io/wordpress-plugin/pull/564) Sweep the paired `_transient_timeout_beyondwords_*` rows on uninstall, so no orphaned option rows are left behind.
--- a/src/editor/classic/class-metabox.php
+++ b/src/editor/classic/class-metabox.php
@@ -97,6 +97,8 @@ class Metabox {
 			} else {
 				self::player_embed( $post );
 			}
+		} else {
+			self::awaiting_generation_embed( $post );
 		}
 
 		// The Embed dropdown ("None" = no player) replaces the old Display player checkbox.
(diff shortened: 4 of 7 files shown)


gouravkhunger:

Soft score: **Approve**.
Correct gating, no regression (else-branch early-returns unless a job is queued + project set), reuses the existing poll infra, PHPUnit added — but the two Cypress specs were not run (wp-env mounted `main`), so run `content-id.cy.js` before merging.

<details>

## Architectural map
Fixes classic-editor metabox showing an **empty Player section** (no player, no spinner) during the *first* post-publish wait — when generation is deferred to cron so no content ID exists yet. The prior poll (#567) only covered the second wait (ID exists, still processing).

PHP renders a loading container when a cron job is queued but no ID yet; JS polls the post's own `wp/v2` meta until cron writes the ID, then hands off to the existing `initMetaboxPlayer()`.

## Affected subsystems
- `src/editor/classic/class-metabox.php` — new `awaiting_generation_embed()`, wired into `render_meta_box_content()` else branch.
- `src/editor/components/content-id/classic-metabox.js` — new `awaitContentId()` + init-path branch.
- Tests: PHPUnit (2 cases), Cypress (2 specs + `scheduleAudioGeneration` task).
- Docs: `doc/async-rest-migration.md`, `readme.txt` changelog.

## Dependency graph
No new deps. Reuses `Sync::GENERATE_AUDIO_CRON_HOOK`, `Meta::get_project_id`, `Urls::get_js_sdk_url`, and the existing `pollContentStatus()`. No new REST route — leans on `beyondwords_content_id` already being `show_in_rest`.

## Risky files
`classic-metabox.js` — most logic, +108 lines. The poll reuses `pollContentStatus` by reporting non-terminal `queued` until the ID lands. Verified: `NON_TERMINAL_STATUSES` includes `queued`, `found` is terminal → resolves correctly. Supersede token, hidden-tab handling and 120s budget all inherited. Clean.

## Public API changes
None. No new endpoints, hooks, meta keys, or options.

## DB / auth / security
No DB schema change. The poll is a same-origin `wp/v2` GET with `X-WP-Nonce`, reading own post meta already exposed via REST — no privilege change, no new surface.

## Deployment / config
None. The Cypress `scheduleAudioGeneration` task queues the cron event an hour out so WP-Cron cannot fire it mid-test — test-only.

## Correctness checks (verified)
- **Regression-safe:** the else branch now calls `awaiting_generation_embed()`, but it early-returns unless a cron job is scheduled AND a `project_id` is set. Non-async no-content posts stay blank as before.
- **Gating correct:** `wp_next_scheduled` prevents a spinner that would never resolve.
- **Handoff:** clears `data-await-content`, sets `data-content-id`, then `initMetaboxPlayer()` still waits for `processed` — avoids a 404 the CDN would cache.
- **`getRestBase()` no-arg** falls back to the `#post_type` input, present on the classic edit screen.
- Tests match the behavior; `Cypress.expose` is a real config-backed helper.

## Minor (non-blocking)
- `phpcs:disable WordPress.WP.EnqueuedResources.NonEnqueuedScript` around the new SDK `<script>` — AGENTS.md rule 1 disfavors `phpcs:ignore`, but this mirrors the existing accepted pattern at line 212 for the same SDK tag. Consistent, not newly introduced.
- Doc says `#566`, PR body says `#567` for the prior poll — trivial ref mismatch.

## Review priority order
1. `classic-metabox.js awaitContentId()` (core logic) — verified sound.
2. `class-metabox.php awaiting_generation_embed()` gating — verified sound.
3. Cypress specs — **author flags these were NOT run** (wp-env mounted `main`, not the branch). Must run before merge.

## Verdict
Solid, well-scoped fix. Correct gating, no regression, reuses existing poll infra, PHPUnit added (714/715, single failure pre-existing and unrelated), changelog + doc present. Only real gate: **run the two Cypress specs before merge** as the author requested.

</details>

galbus:
Thanks — two points addressed, one deliberately not.

**Cypress gate: already satisfied.** The specs have since run in CI on this branch and pass on both PHP 8.0 and 8.5:

```
✓ shows a Generating state while deferred generation is still queued
✓ hands off to the content-status poll once the queued job writes an ID
```

I checked the job logs rather than the green tick, since `cypress-fail-fast` can mark specs pending and still leave a job looking healthy. The PR body carried a stale "have not been run" warning from when I opened it — that's now corrected. (An earlier Plugin Check failure on this PR was a Docker Hub registry timeout during `wp-env start`, not a code issue; it passed on re-run.)

**Ref mismatch: fixed.** You're right, and the direction was the opposite of what I'd assumed — #566 is "poll content status before embedding the player"; #567 is the voice Language/Accent split. `doc/` was correct, the PR body was wrong, and the body is now fixed. Note the same wrong ref is in commit `17fb28c`'s message; I left history alone rather than force-push over a branch you'd already reviewed, but say the word if you'd prefer it amended.

**`phpcs:disable` on the SDK `<script>`: not changed, deliberately.** Agreed it sits awkwardly against AGENTS.md rule 1. The honest fix is `wp_enqueue_script()` rather than a raw tag, but that would mean changing `player_embed()`'s two existing emissions too — and the `player_embed_neutralises_content_id_xss()` test asserts exactly one container and one `<script>` in that output, which moving the tag to the footer would break. That's a wider refactor than this bug fix, so I've kept the existing accepted pattern. Happy to split it out as its own PR if you want it cleaned up properly.

gouravkhunger (approved):

PR #604, opened 2026-07-26
fix: pinned sidebar button aria-label rendered "[object Object]"

galbus:
## Problem

In the block editor top bar, the pinned BeyondWords sidebar button had `aria-label="[object Object]"`. The tooltip and screen reader name were both broken; the button itself worked.

## Root cause

`ComplementaryArea` passes the `PluginSidebar` `title` straight through as the pinned button's `label`:

```js
// @wordpress/interface/src/components/complementary-area/index.js:279
label={ title }
```

…and `Button` resolves that to the accessible name:

```js
// @wordpress/components/src/button/index.tsx:236
'aria-label': additionalProps[ 'aria-label' ] || label,
```

`a1e54eb` changed `title` from a plain string to `<BeyondwordsTitle />` in order to show the brand icon in the sidebar header. React stringifies a non-string DOM attribute value, so the accessible name — and the tooltip, which uses the same value — became `[object Object]`. `PluginSidebar` documents `title` as `{string}`.

## Fix

- `title` is a plain translated string again.
- The header brand mark (the reason `a1e54eb` switched to an element) is restored via a CSS mask on a new `headerClassName="beyondwords-sidebar__header"`.

The `header` prop looked like the tidier fix but is not viable: the pin/unpin star lives *inside* the `header || (…)` fallback ([complementary-area/index.js:323-342](https://github.com/WordPress/gutenberg/blob/trunk/packages/interface/src/components/complementary-area/index.js)), so supplying `header` would silently delete it.

## Reviewer notes

- The mask's SVG path is copied verbatim from `BeyondwordsIcon` — verified byte-identical, and the data URI decodes to well-formed SVG with the correct `xmlns`/`viewBox`. An earlier pass omitted the `data:image/svg+xml,` prefix, which rendered nothing silently; that is fixed and asserted against.
- **Known fragility:** the icon rule hangs off `.interface-complementary-area-header__title`, a WordPress-internal class. If WP renames it the icon disappears with no error and no failing test. The `aria-label` fix itself has no such dependency.
- The other three `<BeyondwordsTitle />` call sites (`addControls`, `prepublish`, `document-setting`) are **not** affected — they feed `PanelBody`, which renders `title` as button children rather than an attribute, so an element is fine there.
- `PluginSidebarMoreMenuItem` is not a duplicate of the menu item `ComplementaryArea` renders itself; it passes `__unstableExplicitMenuItem`, which suppresses the implicit one.

## Testing

- ESLint and Prettier pass; stylelint introduces no new error categories (the file's pre-existing indentation/`color-named` nits are unchanged, and that config is not enforced on this file).
- No Cypress spec is affected — `openBeyondwordsPluginSidebar` opens the sidebar via the "BeyondWords sidebar" link, not the pinned button.
- **Not verified in a running editor.** The mask render and the corrected `aria-label` are structurally verified only. To confirm: `npm run build && npm run env:start`, then hover the pinned BeyondWords icon — tooltip and accessible name should read "BeyondWords", with the brand mark still atop the sidebar header.

Pre-commit hook was skipped (`--no-verify`): it requires `vendor/bin/grumphp`, which is not installed in this worktree. Its `git_pre_commit` suite is entirely PHP-scoped (`phpcs`/`phplint` are `triggered_by: php`; `composer`/`lint_plugins` cover untouched files) and this change is JS + CSS only.


gouravkhunger:

Soft score: **Approve**
- Root cause (element stringified into `aria-label`) correctly diagnosed; fix is minimal and right.
- One thing to confirm before merge: the svgr `ReactComponent` import backs **all** panel icons — verify the built bundle actually renders the icon, since a loader mismatch fails silently.

<details>

## Architecture map
Tiny, focused JS+CSS bugfix. 5 files, +30/−26. No PHP, no DB, no API, no auth, no deps.

**Subsystems touched:** block editor sidebar (`src/editor/block/sidebar`), shared icon component (`src/editor/components/icon`), sidebar CSS.

## Root cause + fix (correct)
`a1e54eb` set `PluginSidebar title={<BeyondwordsTitle/>}`. `ComplementaryArea` forwards `title` as the pinned button `label`, `Button` maps `label` to `aria-label`. React stringified the element = `[object Object]`. Fix reverts `title` to plain string `__('BeyondWords')` and restores the header mark via a CSS mask. Diagnosis sound, fix minimal and right.

## Dependency graph change
`icon/index.js` no longer hand-writes SVG JSX. Now single source `beyondwords.svg`, consumed two ways: JS via `@svgr/webpack` `ReactComponent`, CSS via `url()` mask. Good de-dup — kills the drift risk between the JSX path and the CSS path.

## Risky files (priority order)
1. **`src/editor/components/icon/index.js`** — `import { ReactComponent as BeyondwordsIcon } from './beyondwords.svg'` depends on wp-scripts' svgr+url-loader chain exporting `ReactComponent`. If loader order/config differs from assumed, `BeyondwordsIcon` is `undefined` and every panel icon (sidebar + prepublish + document-setting + addControls all import from here) breaks silently. Author says verified with a real webpack build — confirm the built bundle renders the icon before merge. Highest-blast-radius file.
2. **`src/editor/components/sidebar/sidebar.css`** — mask hangs off `.interface-complementary-area-header__title`, a WP-internal class. Author flagged it; cosmetic-only, silent-disappear failure mode. Acceptable, documented.
3. `block/sidebar/index.js` — the actual fix, low risk.

## Public API changes
`icon/index.js`: `BeyondwordsIcon` now default-exports the svgr component (was a named function). Named + default exports preserved, all 4 call sites still resolve. No break.

## No DB / auth / config / deployment changes.

## Notes (non-blocking)
- No test added. AGENTS rule 2 wants Cypress coverage per JS feature; `aria-label` isn't currently asserted and there's no existing spec exercising the pinned button. Regression could recur unseen — a small assertion on the pinned button's accessible name would close it.
- SVGO drops `focusable="false"` (keeps `aria-hidden="true"`) — fine, WP dropped the browsers that needed it.
- Changelog: one bullet under **Fixes** in the unreleased 7.0.0 block, user-facing wording. Correct per rule 5.
- Comments trimmed to one line each. Compliant with the docs rules.

## Verdict
Clean, correct, well-scoped fix with a strong root-cause writeup. Only real risk is the untested svgr `ReactComponent` build assumption (broad blast radius across all panel icons) — verify in a running build.

</details>

galbus:
Verified the svgr `ReactComponent` assumption end-to-end — it holds, so no code change was needed.

Reinstalled from the post-merge lockfile (`npm ci`, so **svgo 3.3.4** and `@svgr/webpack` 8.1.0 — my earlier build predated the dependabot bump), then compiled the icon module through **wp-scripts' own `module.rules`, unmodified**, and rendered it with `react-dom/server`:

```
{"namedType":"function","defaultType":"function","defaultIsSameAsNamed":true}

--- BeyondwordsIcon ---
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" aria-hidden="true" viewBox="-4 -4 38 38"><path fill="currentColor" d="M29.4 18.59A20.6 20.6 0 0 0 28 15…

--- BeyondwordsTitle ---
<span style="display:inline-flex;align-items:center;gap:8px"><svg …><path fill="currentColor" …
```

So `BeyondwordsIcon` is a real component (not `undefined`), default and named exports are the same reference — all four call sites resolve — and it renders actual `<svg>`/`<path>` markup rather than a URL string. The silent-failure mode you flagged would show as `namedType: "undefined"` here.

The shipped `build/index.js` from `npm run build` carries the same compiled element (`width:24, height:24, fill:"none", aria-hidden:"true", viewBox:"-4 -4 38 38"` + one `fill:"currentColor"` path).

On the non-blocking test note — agreed, and I'd flagged it too. A Cypress assertion on the pinned button's accessible name is the durable guard for this regression, but verifying a new spec needs a wp-env run I can't do right now, so I've deliberately left it rather than push an unverified test. Happy to add it as a follow-up.

gouravkhunger (approved):

PR #606, opened 2026-07-26
fix(sync): recover when two saves race to create audio

galbus:
## The bug

Publishing can fire two overlapping create requests for the same post. Both read an empty `beyondwords_content_id`, both POST, and the API rejects the second:

```json
{"code":422,"message":"Invalid request body","errors":[{"location":"source_id","message":"has already been taken"}]}
```

The losing request stored `#422: source_id has already been taken`, left the content ID empty and never retried, so the post rendered **no player at all** — auto-prepend and shortcode both suppress on an empty content ID — even though the audio existed in BeyondWords and had processed fine.

Seen on a 7.0.0-beta.1 staging site (classic editor, post 3474): the project held content `effef870-…` with processed audio for `source_id 3474`, while the post had the error meta and no content ID. Hit rate in a full matrix run: 1 publish in 18. The only recovery was pasting the content ID into the Data panel and hitting Fetch.

## The fix

Two guards, both explained in [doc/source-id-race.md](https://github.com/beyondwords-io/wordpress-plugin/blob/claude/plugin-v7-source-id-race-21462f/doc/source-id-race.md).

**1. Don't make the second POST.** `Sync::create_audio_once()` takes a per-post lock (`_beyondwords_create_lock`, 30s) before creating and re-reads the content ID under it, in case the winner finished in the meantime. The lock is taken with `add_post_meta( …, $unique: true )`, whose existence check is an *uncached* query, so two racing requests contend over one statement rather than the whole API round trip. The response is stored **inside** the lock — releasing first would leave a window with no content ID, which is exactly the state the lock exists to hide. A lock older than the timeout is stolen, so a request that dies mid-create can't lock a post out of audio permanently.

**2. Adopt what the winner created.** The content endpoint resolves either a content ID or a `source_id`, and the plugin sends the post ID as the source ID. So `Client::create_audio()` treats that 422 as a signal rather than an error: it re-fetches the content that already exists, adopts it, and clears the error. `Sync::process_response()` then stores the content ID and preview token exactly as after a successful create. Matched on the error's `location`, never the message text.

## Reviewer notes

**Adoption is restricted to this site's own content.** Source IDs are bare post IDs and are **not** namespaced by site, so a staging clone or a second subsite on one project collides on every post ID — and this 422 was previously what stopped two installs fighting over one content record. Adoption therefore requires the fetched record's `source_url` to sit under this site's `home_url()`; anything else leaves the 422 stored, i.e. the pre-fix behaviour. Checked against the site root rather than the post's permalink, because the racing saves can straddle the draft → published permalink change.

**Known trade-off in guard 1.** If the lock holder's create fails (a timeout, say), a genuine concurrent save is suppressed and the post waits for the next save. Failures are usually correlated across concurrent requests, so this is narrower than the double-create it prevents — but it is a deliberate choice, not an oversight.

**Not atomic, by necessity.** The acquire is near-atomic, and on a host routing reads to a replica neither request may see the other's row. Guard 2 absorbs both. A truly atomic lock needs direct SQL, which the VIP-Go ruleset forbids.

**Uninstall.** `Uninstaller::cleanup_custom_fields()` also sweeps the lock key, since a request that fatals mid-create can strand a row.

## Verified against the live API

Before implementing, confirmed on project 9969 that `GET /projects/{id}/content/{source_id}` resolves a source ID (returning the record with its real `id`, `preview_token` and `source_url`), 404s for an unknown one, and that a duplicate create returns the 422 body above. A rejected create makes nothing, so the probe was side-effect free.

## Testing

11 new PHPUnit tests under `@group source-id-race`, covering adoption, the cross-install refusal, an unrelated 422, a failed lookup, and each lock behaviour — including a regression test that the content ID is stored *before* the lock lifts (confirmed failing against the earlier ordering).

Full suite green: **723 tests, 0 failures**, repeatable. `phpcs` clean on all three source files under `WordPress-VIP-Go` with no new `phpcs:ignore`. Coverage wasn't measurable locally (no xdebug/pcov in the tests container) — CI will report it.

## Follow-up spotted, not fixed here

`Sync::on_add_or_update_post()` discards the return of `delete_audio_for_post()`, so a failed "Remove audio" DELETE (timeout, 401) still wipes all local meta and orphans the remote content. Out of scope for this PR.


gouravkhunger:

Soft score: **Approve**

- Guards defend each other correctly; content ID lands **before** the lock lifts (`finally`-released, regression-tested).
- Cross-site adoption gate (`source_url` under `home_url()`) is the right security call and fails safe.
- Minor, non-blocking: `create_audio_once()` returning `false` when lock-held can miscount a post in a concurrent **bulk** run (`class-sync.php:407`) — cosmetic, not a data bug.

<details>

## What it does
Fixes race where two overlapping publish requests both POST create for one post. Second gets `422 source_id has already been taken`, stored the error with an empty content ID, and the post rendered no player. Two guards: (1) a per-post create lock so the second POST never fires; (2) on 422, re-fetch the existing content and adopt it.

## Affected subsystems
- `src/post/class-sync.php` — new `create_audio_once()` + `acquire_create_lock()`, lock consts.
- `src/api/class-client.php` — `adopt_existing_content()` + `is_duplicate_source_id()`.
- `src/core/class-uninstaller.php` — sweeps the new lock meta key.
- Docs (`doc/source-id-race.md`, README, changelog), tests (3 phpunit files).

## Dependency graph
`generate_audio_for_post()` now delegates the create branch to `create_audio_once()` (new indirection; stores its own response under the lock). `create_audio()` gains a post-POST adoption hook. Uninstaller depends on `Sync::CREATE_LOCK_META_KEY`. No new external deps.

## Risky files
`class-sync.php` highest — concurrency logic, lock lifecycle, response-store ordering. `class-client.php` second — new API round-trip + cross-site guard on the 422 path.

## Public API changes
None external. New private methods + two `const` on `Sync`. `generate_audio_for_post()` signature unchanged.

## DB changes
New post meta `_beyondwords_create_lock` (underscore-prefixed, uncached `add_post_meta $unique`). Swept on uninstall. No schema/migration.

## Auth / security
Adoption gated on the fetched `source_url` starting with `trailingslashit( home_url() )` — blocks a staging clone / sibling subsite on the same project from stealing content. Correct and important; fails safe (keeps the 422) if unconfirmed. No nonce/escaping surface touched.

## Deploy / config
None. Pure runtime behaviour. VIP-Go clean (no direct SQL — the lock is meta-based by necessity, acknowledged non-atomic).

## Correctness notes
- **Ordering correct:** `process_response()` runs inside `try`, `delete_post_meta` on the lock in `finally` — the content ID lands before the lock lifts. Regression-tested. ✅
- **Lock steal:** a stale lock (> 30s) is stolen via `update_post_meta`, so a dead mid-create request can't lock a post out forever. ✅
- **Non-atomic acquire:** `add_post_meta($unique)` is SELECT-then-INSERT, not atomic; two racers can both win. Explicitly acknowledged and absorbed by guard 2. Acceptable given VIP forbids direct SQL. ✅

## Minor / non-blocking
- `create_audio_once()` returns `false` when the lock is held or content already exists. `generate_audio_for_post()` returns that up. The bulk counter (`class-sync.php:407`) treats falsy as failure, so a concurrent-covered post could be miscounted in a bulk run — rare, cosmetic, not a data bug.
- `home_url()` match is scheme/host-exact: an `http`↔`https` or `www` mismatch in a stored `source_url` would decline adoption. Fails safe (422 kept), so acceptable, but worth noting if source URLs drift.
- `is_duplicate_source_id` matches any 422 with `location=source_id` (not just "already taken"). A malformed-source_id 422 triggers one extra GET that 404s and falls through — harmless, documented.

## Testing
11 new tests under `@group source-id-race`: adoption, cross-install refusal, unrelated 422, failed lookup, lock-held, lock-steal, and a store-before-unlock regression. Good coverage of the concurrency contract. Full suite reported green (723). Coverage deferred to CI.

## Review priority order
1. `class-sync.php` lock lifecycle + response ordering.
2. `class-client.php` adoption + cross-site guard.
3. Tests.
4. Uninstaller / docs.

## Verdict
**Approve.** Well-scoped fix, the guards defend each other, the security gate on adoption is the right call, tests are thorough, and the non-atomic trade-off is stated honestly. The minor bulk-count semantics are worth a glance but not blocking.

</details>

galbus:
Thanks — acted on one of the three, pushed as 8dc98a4.

**Scheme-exact `home_url()` match** — fixed. You're right that this one bites: a site moved from `http` to `https` stores the old scheme in the `source_url` of everything it created before the move, so the exact-prefix match declined its own content and left the 422 in place — precisely the bug this branch exists to fix. The comparison now strips the scheme from both sides. Host and path still have to match, so a staging clone or a sibling subsite (`example.com/site-a` vs `example.com/site-b`) is still refused, and an empty site root now declines rather than matching everything. Covered by a new test and noted in `doc/source-id-race.md`.

**Bulk counter miscount** (`class-sync.php:407`) — not changing it here, deliberately. The counter treats every falsy return as a failure, so this predates the PR in shape: `BEYONDWORDS_AUTOREGENERATE` being off already counts a skipped post as failed. Fixing it honestly means a fourth `skipped` bucket threaded through `bulk_generate_audio_for_posts()`, the AJAX handler and the admin notice, plus their tests — a wider change than this fix should carry, and the miscount is cosmetic (re-running the action is idempotent and would then report success). Logged as a follow-up.

**`is_duplicate_source_id` matching any `location=source_id` 422** — no change, this is the intended design. Matching the `location` rather than the free-form message is deliberate, and the fallthrough you describe is the safe branch: a malformed-`source_id` 422 costs one GET that 404s, and the original error stands.

gouravkhunger (approved):

PR #610, opened 2026-07-28
fix: count skipped posts separately from failures in bulk generate audio

galbus:
## The bug

`Sync::bulk_generate_audio_for_posts()` counted the outcome of each post like this:

```php
if ( self::generate_audio_for_post( $post_id ) ) {
    ++$generated;
} else {
    ++$failed;
}
```

Every falsy return was a failure — but several of `generate_audio_for_post()`'s falsy returns mean *nothing to do*:

- `BEYONDWORDS_AUTOREGENERATE` is defined false and the post already has a content ID.
- `should_generate_audio_for_post()` returns false (autosave, revision, ineligible post status).
- `create_audio_once()` returns false because another request holds the per-post create lock, or already stored a content ID — added in #606 and raised there as minor/non-blocking.

So bulk "Generate audio" over, say, a set of drafts reported *"3 posts failed, check for errors in the BeyondWords column below."* when nothing had gone wrong and there were no errors to find. Cosmetic — re-running the action is idempotent — but the count was wrong and pointed the user at a column with nothing in it.

## The fix

`generate_audio_for_post()`'s falsy return can't carry the distinction, and changing its return type would ripple into `on_add_or_update_post()` (which casts to bool) and the cron callback. So the pipeline moved into a new private `Sync::generate_audio_result()` that returns the outcome alongside the response:

```php
[ 'outcome' => self::OUTCOME_SKIPPED, 'response' => false ]
```

`generate_audio_for_post()` is now a one-line wrapper returning `['response']` — **its public contract is unchanged**, and no existing caller was touched. `create_audio_once()` returns the same shape, so both of #606's skip returns are covered. `bulk_generate_audio_for_posts()` returns a fourth `skipped` count.

`BulkEdit::handle_bulk_generate_action()` adds `beyondwords_bulk_skipped` to the redirect when non-zero (mirroring how `deferred` is handled), and both bulk handlers strip the arg on redirect. `Notices::skipped_notice()` renders it:

> 3 posts were skipped because no audio change was needed.

`notice-info`, since nothing needs acting on.

## Notes for review

- **The AJAX handler needed no functional change.** `BulkEdit::save_bulk_edit()` calls `BulkEdit::generate_audio_for_posts()`, which only flags posts for the next save and never produces counts. It gets one boundary test documenting that, rather than invented coverage.
- **The skipped message is deliberately one message for three causes** (ineligible status / autoregenerate off / concurrent create). Splitting it would mean three query args and three notices for a cosmetic count; happy to if you'd rather.
- Rationale for the outcome split lives in [doc/async-rest-migration.md](doc/async-rest-migration.md) next to the existing bulk-generation section; the code carries a one-line pointer.
- Also fixed a drifted translator comment on `failed_notice()` that read "number of posts that were skipped" — now genuinely confusing next to a real skipped notice.

## Testing

- `composer test` — 746 pass, 1 skipped (the pre-existing multisite skip). PHPCS `WordPress-VIP-Go` clean, no `phpcs:ignore` added.
- New PHPUnit coverage: drafts and a held create lock both count as `skipped` (both assert via `pre_http_request` that no API call is made); the redirect carries `beyondwords_bulk_skipped` and omits it when zero, including dropping a stale value from the incoming URL; the notice's markup and both plural forms.
- One test also asserts `generate_audio_for_post()` still returns falsy for a skipped post, pinning the unchanged contract that `on_add_or_update_post()` depends on.

**Two gaps to flag:**

1. **The Cypress case is unverified.** `bulk-actions.cy.js` gains a per-post-type test that bulk-generates over a draft and asserts the skipped notice appears and the failed notice does not — the exact user-visible bug, and the only skip cause reachable from the UI. I couldn't run it locally: wp-env serves the plugin from the main checkout, so a run from the worktree would exercise `main`'s code rather than this branch's. The existing assertions in that spec are unaffected (nothing in those flows skips). Worth a CI eye.
2. **The coverage gate didn't run locally** — no xdebug/pcov in the tests container, so `coverage-check` had no clover.xml. CI will compute it.

### DIFF
--- a/doc/async-rest-migration.md
+++ b/doc/async-rest-migration.md
@@ -110,6 +110,15 @@ It sets `beyondwords_generate_audio` meta on every selected post, then:
   ([src/posts-list/class-notices.php](../src/posts-list/class-notices.php));
   their generate flag is already set, so re-running the action completes them.
 
+Each inline generate is counted as `generated`, `skipped` or `failed`. Skipped
+means there was nothing to do, not that anything went wrong: an ineligible post
+status, an existing recording with `BEYONDWORDS_AUTOREGENERATE` off, or another
+request already creating the audio ([source-id-race.md](source-id-race.md)).
+`Sync::generate_audio_result()` carries that outcome alongside the API response,
+because a falsy response alone can't tell the two apart —
+`Sync::generate_audio_for_post()` still returns just the response, so the save
+and cron callers are unaffected. `Notices::skipped_notice()` surfaces the count.
+
 ## Known limitations
 
 ### Classic-editor server-side reads
--- a/readme.txt
+++ b/readme.txt
@@ -111,6 +111,7 @@ Release date: tbc
 
 **Fixes**
 
+* [#610](https://github.com/beyondwords-io/wordpress-plugin/pull/610) Stop the bulk "Generate audio" notice counting skipped posts as failures.
 * [#606](https://github.com/beyondwords-io/wordpress-plugin/pull/606) Fix posts published with no player when two saves race to create audio.
 * [#564](https://github.com/beyondwords-io/wordpress-plugin/pull/564) Sweep the paired `_transient_timeout_beyondwords_*` rows on uninstall, so no orphaned option rows are left behind.
 * [#540](https://github.com/beyondwords-io/wordpress-plugin/pull/540) Escape the player `onload` attribute to prevent stored XSS via the Content ID.
--- a/src/post/class-sync.php
+++ b/src/post/class-sync.php
@@ -48,6 +48,15 @@ class Sync {
 	const BULK_GENERATE_SYNC_LIMIT = 10;
 
 	/**
+	 * Outcome of one generate attempt: audio requested, nothing to do, or the API call failed.
+	 *
+	 * @since 7.0.0
+	 */
+	const OUTCOME_GENERATED = 'generated';
+	const OUTCOME_SKIPPED   = 'skipped';
+	const OUTCOME_FAILED    = 'failed';
+
+	/**
 	 * Post meta key for the per-post "a create is in flight" lock.
 	 *
 	 * Underscore-prefixed so core protects it: request bookkeeping, not post data.
@@ -216,13 +225,27 @@ class Sync {
 	 * @return array<mixed>|false|null Response from the API, or false when audio wasn't generated.
 	 */
 	public static function generate_audio_for_post( int $post_id ): array|false|null {
+		return self::generate_audio_result( $post_id )['response'];
+	}
+
+	/**
+	 * Generate audio for a post, reporting whether it ran, was skipped, or failed.
+	 *
+	 * A falsy response on its own can't tell a post that had nothing to do from
+	 * one whose API call failed. See doc/async-rest-migration.md.
+	 *
+	 * @since 7.0.0
+	 *
+	 * @return array{outcome:string, response:array<mixed>|false|null}
+	 */
+	private static function generate_audio_result( int $post_id ): array {
 		if ( ! self::should_generate_audio_for_post( $post_id ) ) {
-			return false;
+			return self::skipped_result();
 		}
 
 		$post = get_post( $post_id );
 		if ( ! $post ) {
-			return false;
+			return self::skipped_result();
 		}
 
 		$integration_method = \BeyondWords\Settings\Fields::get_integration_method( $post );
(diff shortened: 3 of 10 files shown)


gouravkhunger:

Soft score: **Approve**
- Public `generate_audio_for_post()` contract preserved and pinned by test; `skipped` correctly diverted before counting; nonce/escape reused correctly in the new notice.
- Only follow-up: confirm CI green (Cypress + coverage were deferred locally, as you flagged).

<details>

## What
Bulk "Generate audio" counted every falsy `generate_audio_for_post()` return as *failed*. But several falsy returns mean *nothing to do* (ineligible status, autoregenerate off with existing content ID, lost create-lock from #606), producing a false "N posts failed" notice pointing users at an empty error column. This PR splits `skipped` out from `failed`.

## Affected subsystems
- `src/post/class-sync.php` — new outcome-carrying pipeline (`generate_audio_result`, `skipped_result`, `attempted_result`), fourth `skipped` count in bulk.
- `src/posts-list/class-bulk-edit.php` — redirect arg `beyondwords_bulk_skipped` add + strip.
- `src/posts-list/class-notices.php` — new `skipped_notice()`; fixed drifted translator comment on `failed_notice()`.
- Tests: PHPUnit (sync, bulk-edit, bulk-edit-ajax, notices) + Cypress bulk-actions.
- Docs: `doc/async-rest-migration.md`, `readme.txt` changelog.

## Dependency graph
No new externals. Internal restructure only: `generate_audio_for_post()` now delegates to private `generate_audio_result()`. `create_audio_once()` return type narrowed `array|null|false` → `array` (private, single caller). Clean.

## Public API changes
None. `generate_audio_for_post()` signature and falsy contract preserved — pinned by test asserting `on_add_or_update_post()`/cron callers unaffected. New `OUTCOME_*` consts and `skipped_notice()` are additive.

## DB / auth / config / deploy
- DB: none. No schema, no new option/meta keys.
- Auth: `skipped_notice()` reuses `get_query_count()` → `verify_result_nonce()`, sanitizes count, escapes output. Same proven pattern as sibling notices. No gap.
- Config/deploy: none.

## Correctness check
Logic verified end-to-end. `attempted_result($response)` maps truthy→generated, falsy→failed — same as old behavior for real API calls. Skip returns now diverted before counting. Early return `create_audio_once()` still calls `process_response` internally, no double-processing. Both bulk handlers strip stale `beyondwords_bulk_skipped` on redirect (test confirms drop of stale incoming value). Correct.

## Risky files
`class-sync.php` is the hot path (every save + cron + bulk). Change is a pure refactor of counting; single-post contract explicitly tested. Low risk.

## Notes / minor
- One `skipped` message covers three distinct causes. Acceptable — flagged in the description; splitting would add 3 args/notices for a cosmetic count.
- Client-side (`get_player_by_source_id`) falsy still counts as `failed`, not `skipped` — unchanged from prior behavior, not a regression.
- Two flagged gaps: Cypress unverified locally (wp-env serves main), coverage gate deferred to CI. Both legitimate CI-side; worth a green-run confirmation.

## Review priority order
1. `src/post/class-sync.php` (core logic) — verified.
2. `src/posts-list/class-notices.php` (nonce/escape) — verified.
3. `src/posts-list/class-bulk-edit.php` (redirect arg lifecycle) — verified.
4. Tests — thorough: skip via draft + locked create, redirect present/absent/stale-dropped, notice markup + both plurals, AJAX boundary.

## Verdict
Well-scoped, well-tested bug fix. Contract preserved, security patterns reused correctly, no DB/API impact. Confirm CI (Cypress + coverage) green before merge.

</details>

galbus:
No code changes — the only follow-up was the CI confirmation, so here it is: all checks green on `f3f7eb4`, including both gaps flagged in the description.

**Cypress** — pass on PHP 8.0 (34m36s) and 8.5 (36m8s). The new case ran on both, across all three post types:

```
✓ reports skipped, not failed, for draft Posts (6269ms)
✓ reports skipped, not failed, for draft Pages (6260ms)
✓ reports skipped, not failed, for draft CPT Actives (8481ms)
```

That is the end-to-end proof I could not get locally (wp-env serves the plugin from the main checkout, so a worktree run would exercise `main`): a real bulk "Generate audio" over a draft now renders the skipped notice and no failed notice.

**Coverage** — `composer test:coverage-check` reports `Total code coverage is 91.71 % – OK!`, against the 85% gate. PHPUnit passes on 8.0 and 8.5.

Also green: Plugin Check, GrumPHP, ESLint (wp-prettier), Jest, Build Plugin ZIP.

gouravkhunger (approved):

PR #611, opened 2026-07-28
Tidy classic metabox divider spacing and Content ID layout

galbus:
Two CSS fixes to the classic editor metabox, from a review of the rendered sidebar.

## What changed

**Section dividers.** The `<hr />` rules between sections relied on the browser default margin, so they sat tighter than the 8–13px rhythm of the headings and fields around them. They now carry a `beyondwords-metabox__divider` class with a 16px symmetric margin — measured 16px above and below each rule in the rendered metabox.

**Content ID / Fetch.** The field and button shared a flex row, which squeezed the field on a narrow sidebar (the Content ID UUID was clipped mid-value). The field is now full width, with the button on its own right-aligned row 9px below it.

Two details worth knowing:

* Core gives inputs a 1px side margin, which would push a `width: 100%` field 1px past the edge the button aligns to — zeroed on the field so the two share a right edge.
* The fetch spinner now goes in *before* the button rather than after it, and is un-floated (core floats spinners right). The button therefore holds the same right edge whether idle or loading, instead of jumping 24px left when a fetch starts.

**`classic.css` reformatted from spaces to tabs.** The file failed `npm run lint:css` on every line before this change; `wp-scripts lint-style --fix` converted it, which is what the WordPress standard and the AGENTS.md "reformat surrounding code when you touch it" rule ask for. This inflates the diff — the only substantive additions are the four new rule blocks at the end. The one remaining lint error (`#speechkit_status`, line 7) is pre-existing and unfixable: it's a WordPress-generated column id.

## Testing

* `phpcs` (WordPress-VIP-Go) clean on both changed PHP files.
* `lint:css` clean apart from the pre-existing `#speechkit_status` error.
* `ContentIdTest` (13 tests) and `MetaboxTest` (15 tests) pass.
* Layout verified in a browser against the real `wp-admin` stylesheets, in both the idle and loading states, with computed geometry checked (field spans the full content box; field and button right edges align to 0px in both states).
* Jest suite passes (52 tests); `lint:js` clean on the changed JS.

`tests/cypress/e2e/classic-editor/content-id.cy.js` was not run locally — the dev wp-env serves the main checkout rather than this worktree, so it would have exercised the pre-change code. The spec targets only ids this change leaves intact (`#beyondwords-metabox-content-id`, `#beyondwords_content_id`, `#beyondwords__content-id--fetch`), so CI covers it.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -108,6 +108,7 @@ Release date: tbc
     * Editor dropdown data is cached in 15-minute transients; on VIP, audio create/update is deferred to WP-Cron so the save request returns immediately.
 * [#598](https://github.com/beyondwords-io/wordpress-plugin/pull/598) Send taxonomy terms as `tags`.
     * Breaking: the `metadata` param is no longer sent. Code hooking `beyondwords_content_params` to write `$params['metadata']` must append to `$params['tags']` instead, or it will raise a fatal error.
+* [#611](https://github.com/beyondwords-io/wordpress-plugin/pull/611) Tidy the classic editor metabox spacing and Content ID layout.
 
 **Fixes**
 
--- a/src/editor/classic/class-metabox.php
+++ b/src/editor/classic/class-metabox.php
@@ -105,23 +105,23 @@ class Metabox {
 		\BeyondWords\Editor\Components\SettingsFields::render_player_section( $post );
 		( new \BeyondWords\Editor\Components\GenerateAudio() )::element( $post );
 
-		echo '<hr />';
+		echo '<hr class="beyondwords-metabox__divider" />';
 		self::heading( __( 'Content', 'speechkit' ) );
 		\BeyondWords\Editor\Components\SettingsFields::render_content_section( $post );
 
-		echo '<hr />';
+		echo '<hr class="beyondwords-metabox__divider" />';
 		self::heading( __( 'Format', 'speechkit' ) );
 		\BeyondWords\Editor\Components\SettingsFields::render_format_section( $post );
 
-		echo '<hr />';
+		echo '<hr class="beyondwords-metabox__divider" />';
 		self::heading( __( 'Voice', 'speechkit' ) );
 		( new \BeyondWords\Editor\Components\SelectVoice() )::element( $post );
 
-		echo '<hr />';
+		echo '<hr class="beyondwords-metabox__divider" />';
 		self::heading( __( 'Data', 'speechkit' ) );
 		\BeyondWords\Editor\Components\ContentId::element( $post );
 
-		echo '<hr />';
+		echo '<hr class="beyondwords-metabox__divider" />';
 		self::help();
 	}
 
--- a/src/editor/classic/classic.css
+++ b/src/editor/classic/classic.css
@@ -70,9 +70,9 @@ table tr th#speechkit_status {
 	text-align: right;
 }
 
-/* Keeps the button flush right while the floated core spinner is shown. */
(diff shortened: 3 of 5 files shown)


gouravkhunger:

Soft score: **Approve**.
Correct, self-contained classic-metabox CSS/markup change; scope verified (`ContentId::element()` and new classes touch classic editor only, block editor untouched), all Cypress-targeted ids preserved, escaping intact.

<details>

## Architectural map
Pure presentation change to the **classic editor metabox** only. Two visual fixes: `<hr>` dividers get a 16px symmetric margin via a new `beyondwords-metabox__divider` class; the Content ID field becomes full-width with the Fetch button on its own right-aligned row. The bulk of the diff is a `classic.css` space→tab reformat (harmless noise).

## Affected subsystems
- `src/editor/classic/` — metabox container markup + stylesheet.
- `src/editor/components/content-id/` — `ContentId::element()` markup.
- `readme.txt` — changelog bullet.

Scope confirmed: `ContentId::element()` is called only from the classic metabox (`class-metabox.php:122`); the new CSS classes exist only in `classic.css`. The block editor uses its own React `index.js`, untouched.

## Dependency graph changes
None. No new classes, hooks, or imports. The inline `style="margin:8px 0 13px"` was moved verbatim into a `#beyondwords-metabox-content-id` CSS rule — equivalent.

## Risky files
None material. `classic-metabox.js` is unchanged; the spinner is still injected as the button's next sibling (`insertBefore(s, button.nextSibling)`, line 603) — it now lands inside the `.actions` div, and the new `.actions .spinner { float: none }` rule un-floats it. Consistent.

## Public API / DB / auth / security / config
No API, DB, auth, or config changes. All ids the Cypress spec targets (`#beyondwords_content_id`, `#beyondwords__content-id--fetch`, `#beyondwords-metabox-content-id`) are preserved. Escaping unchanged (`esc_attr` intact). No nonce/security surface touched.

## Minor notes (not blocking)
- While loading, the spinner sits to the *right* of the button in the right-aligned actions row, so the button isn't strictly flush-right during load — cosmetic only, slightly contradicts the PR wording.
- The changelog bullet landed under **Enhancements**; the clipped-UUID part reads more like a **Fix**. Borderline, reviewer's call.
- Cypress `content-id.cy.js` was not run locally (worktree/wp-env mismatch) — CI covers it; acceptable given the ids are unchanged.

## Review priority order
1. `class-content-id.php` markup (flex row → stacked) — verified correct.
2. `classic.css` new four rule blocks — verified; the rest is reformat.
3. `class-metabox.php` divider class — trivial.

**Verdict:** low risk, self-contained, correct. Approve.

</details>

galbus:
Thanks — the three minor notes, in order:

**Spinner / button not flush right during load.** Fixed in e10dacc, rather than just re-wording. The spinner is now inserted *before* the button instead of after it, so the button holds the same right edge in both states (measured 0px from the field's right edge idle and loading) instead of jumping 24px left when a fetch starts. The CSS comment is reworded to state what the rule actually does — un-float, because core floats spinners right — rather than claiming it keeps the button in place.

**Changelog heading.** Leaving it under **Enhancements**. The flex row that squeezed the field was introduced in c22d6c1, which no tag contains — it's part of the unreleased v7 editor redesign (#527), so no released version ever clipped the UUID. A **Fixes** bullet would tell users to look for a bug they could never have hit.

**Cypress.** Both jobs have since passed in CI (PHP 8.0 and 8.5), along with the rest of the suite.

gouravkhunger (approved):

PR #598, opened 2026-07-26
feat: send taxonomy terms as a flat `tags` array

galbus:
The content endpoint now expects `tags`, a flat array of strings, in place of the nested `metadata.taxonomy` object. The plugin was still sending the old shape.

## Changes

**[`src/post/class-content.php`](https://github.com/beyondwords-io/wordpress-plugin/blob/claude/wordpress-taxonomy-tags-e8dde6/src/post/class-content.php)**

- The `metadata` key is dropped from the content payload and replaced by `tags`.
- `Content::get_metadata()` and `Content::get_all_taxonomies_and_terms()` are replaced by a single `Content::get_tags()`, which merges the term names from every taxonomy registered for the post type into one flat, deduplicated `string[]`.

A post in category "News" with tag "Audio" now sends:

```json
"tags": ["News", "Audio"]
```

instead of:

```json
"metadata": { "taxonomy": { "category": ["News"], "post_tag": ["Audio"] } }
```

Two deliberate calls:

- **All taxonomies, not just categories.** The old code sent every taxonomy's terms, so keeping that scope means nothing silently stops reaching the dashboard. Only the taxonomy grouping is lost, since the new param is a flat string list.
- **`tags` is always sent, empty array included.** Omitting it when a post has no terms would leave stale tags on the BeyondWords content after someone removes the last term.

**[`tests/phpunit/post/test-content.php`](https://github.com/beyondwords-io/wordpress-plugin/blob/claude/wordpress-taxonomy-tags-e8dde6/tests/phpunit/post/test-content.php)**

- `get_content_params` asserts `tags` is `["Uncategorized"]` and that no `metadata` key is present.
- The `beyondwords_content_params` filter test appends to `$params['tags']` (it previously set `$params['metadata']->custom`).
- The two old metadata tests are replaced by four `@group tags` tests: the default category, a flat merge across a flat + hierarchical custom taxonomy (including a JSON-encodes-as-array assertion), name deduplication across taxonomies, and the empty-array case for a post type with no taxonomies.

## ⚠️ Breaking change

[...]

PR #612, opened 2026-07-28
fix: decode HTML entities in tag names

galbus:
Follow-up to #598. A term named `R&D` reached the dashboard as `r&amp;d`:

## It's our side, not a dashboard rendering bug

WordPress encodes term names **on insert** — `_wp_specialchars` is attached to the `pre_term_name` filter in core's `default-filters.php` — so `R&D` is *stored in the database* as `R&amp;D`. Verified against a real WP install:

```
$ wp eval 'wp_insert_term( "R&D", "post_tag" ); ...'
stored name:   R&amp;D
raw db value:  'R&amp;D'
```

`get_the_terms()` returns that stored value verbatim, so we were sending `R&amp;D` in the `tags` payload. The dashboard then HTML-escaped it for display (correctly — escaping untrusted input is right), which renders the literal text `r&amp;d`. Garbage in, garbage faithfully displayed.

This predates #598 — the old `metadata.taxonomy` payload had the same encoding — but the new `tags` chips make it visible.

## Fix

```php
// Core stores term names HTML-encoded, so "R&D" would reach the API as "R&amp;D".
$tags = array_map( fn( $tag ) => wp_specialchars_decode( $tag, ENT_QUOTES ), $tags );
```

`wp_specialchars_decode()` rather than `html_entity_decode()` because it is the exact inverse of what core applied. A term whose name genuinely contains `&copy;` is stored as `&amp;copy;` and decodes back to the literal `&copy;` the author typed — `html_entity_decode()` would wrongly turn it into `©`. Decoding happens before the dedup pass, so `R&D` and a stray pre-encoded `R&amp;D` collapse to one tag.

## Tests

Both requested cases, in the `@group tags` group:

- **`get_tags_decodes_html_entities_in_term_names`** — asserts the core behaviour being compensated for (`get_term()->name === 'R&amp;D'`), then that `get_tags()` returns `R&D`. The first assertion pins the premise, so if core ever stops encoding, the test says why the decode exists.
- **`get_tags_preserves_utf8_accents_in_term_names`** — `café culture` is stored as raw UTF-8 (`c3a9`) and was never affected, but is now pinned, including a `wp_json_encode()` round trip since the request body escapes the accent as `é`.

Full suite: 717 tests, 2306 assertions, 1 skipped (the pre-existing multisite-only test). `phpcs` clean.

## Not fixed here

`display_name` gets the same treatment from core (`pre_user_display_name` is on the same filter list), so an author called `Smith & Sons` is sent as `Smith &amp; Sons` in the `author` param. Confirmed on a real install, but it's a different payload field and out of scope for this PR.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -111,6 +111,7 @@ Release date: tbc
 
 **Fixes**
 
+* [#612](https://github.com/beyondwords-io/wordpress-plugin/pull/612) Fix tags with an ampersand showing as `r&amp;d` in the BeyondWords dashboard.
 * [#606](https://github.com/beyondwords-io/wordpress-plugin/pull/606) Fix posts published with no player when two saves race to create audio.
 * [#564](https://github.com/beyondwords-io/wordpress-plugin/pull/564) Sweep the paired `_transient_timeout_beyondwords_*` rows on uninstall, so no orphaned option rows are left behind.
 * [#540](https://github.com/beyondwords-io/wordpress-plugin/pull/540) Escape the player `onload` attribute to prevent stored XSS via the Content ID.
--- a/src/post/class-content.php
+++ b/src/post/class-content.php
@@ -406,6 +406,9 @@ class Content {
 			}
 		}
 
+		// Core stores term names HTML-encoded, so "R&D" would reach the API as "R&amp;D".
+		$tags = array_map( fn( $tag ) => wp_specialchars_decode( $tag, ENT_QUOTES ), $tags );
+
 		// Terms in different taxonomies can share a name, and the API wants each tag once.
 		return array_values( array_unique( $tags ) );
 	}
--- a/tests/phpunit/post/test-content.php
+++ b/tests/phpunit/post/test-content.php
@@ -813,6 +813,64 @@ class ContentTest extends TestCase
      *
      * @group tags
      **/
+    public function get_tags_decodes_html_entities_in_term_names()
+    {
+        $taxonomy = 'ampersands';
+
+        register_taxonomy($taxonomy, 'post');
+
+        $term = wp_insert_term('R&D', $taxonomy);
+
+        // Core encodes term names on insert, which is what get_tags() undoes.
+        $this->assertSame('R&amp;D', get_term($term['term_id'])->name);
+
+        $postId = self::factory()->post->create([
+            'post_title' => 'Testing Content::get_tags() with an ampersand',
+            'post_status' => 'publish'
+        ]);
+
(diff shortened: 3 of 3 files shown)


gouravkhunger:

Soft score: **Approve**
Correct fix: `wp_specialchars_decode(..., ENT_QUOTES)` is the exact inverse of core’s `pre_term_name` encoding, decode-before-dedup ordering is right, and both entity + UTF-8 cases are pinned by tests.

<details>

## Map
Follow-up to #598. 3 files, +62/-0 — one production line, rest tests + changelog.

## Affected subsystems
Post domain only — `BeyondWords\Post\Content::get_tags()` (`src/post/class-content.php:410`). Feeds the `tags` param to the outbound BeyondWords API. Nothing else touched.

## What it does
Core encodes term names on insert (`pre_term_name` runs `_wp_specialchars`), so `R&D` is stored as `R&amp;D`. `get_the_terms()` returns the encoded value verbatim, the dashboard re-escaped it and rendered `r&amp;d`. Fix decodes with `wp_specialchars_decode( $tag, ENT_QUOTES )` before the dedup pass.

## Correctness
`wp_specialchars_decode` is the exact inverse of the core encode — correct over `html_entity_decode`, which would wrongly turn a stored `&amp;copy;` into `©` instead of the literal `&copy;` the author typed. `ENT_QUOTES` matches the core flag set. Decode-before-dedup is right: a pre-encoded stray `R&amp;D` collapses with `R&D`.

## Risk
Very low. Pure in-memory string transform, idempotent. No DB/auth/config/API-surface change. Blast radius = tag payload string values only.

## Public API / DB / auth / security / deploy
No internal API change. Outbound payload now sends decoded names — the intended fix. No DB queries, nonces, options, or build/config changes.

## Tests
Two new, `@group tags`:
- entities case pins the premise (`get_term()->name === "R&amp;D"`) then asserts `get_tags()` returns `R&D` — self-documents why the decode exists.
- utf8 case pins `café culture` as unaffected plus a `wp_json_encode()` round trip.

Solid coverage of both the fix and the non-regression.

## Minor / non-blocking
- Description correctly flags `display_name` (author param) as having the same core encoding, left out of scope.
- Test file spacing/`camelCase` matches existing style in `test-content.php` — not new drift.

## Priority order
1. `src/post/class-content.php:409-413` — the one production line. Correct.
2. Tests — assertions sound.
3. `readme.txt` changelog — present, under **Fixes**, one line.

## Verdict
Approve. Tight, correct, well-tested, minimal.

</details>

gouravkhunger (approved):

PR #621, opened 2026-08-12
Prevent a fatal error when the plugin is replaced mid-request

galbus:
## Why

Replacing 7.0.0 with 6.x from the Plugins screen produces a white screen and a fatal-error email, even though the install itself succeeds:

```
Warning: include(.../speechkit/vendor/composer/../../src/core/class-urls.php): Failed to open stream: No such file or directory
Fatal error: Uncaught Error: Class "BeyondWords\Core\Urls" not found in .../src/api/class-client.php:93
```

`Plugin_Upgrader` deletes the installed directory and writes the replacement **within a request that already loaded the old version**. Our hooks are still registered, and `upgrader_process_complete` runs `wp_update_plugins()` — so `http_request_args` fires after the files backing our classes have gone, and `Client::filter_http_request_args()` autoloads `Urls` from a path that no longer exists.

This is new in 7.0.0: v6.x registers no hooks from `ApiClient`, so a running 6.x never touches the autoloader mid-upgrade.

## What changed

`Client::init()` resolves `Urls::get_api_url()` once at bootstrap, while the files are guaranteed to exist, and the filter reads that cached value. Loading `Urls` early also covers `Settings::maybe_print_missing_creds_warning()`, which reaches for `Urls::get_dashboard_url()` from `admin_notices` — rendered on `update.php`.

Rationale and the general rule are in [doc/plugin-upgrades.md](doc/plugin-upgrades.md).

## Note

This can only protect *later* downgrades. The code running during an upgrade is the version being replaced, so a downgrade **from** an already-released 7.0.0 still crashes — nothing shipped afterwards can change that.

## Tests

* `init_loads_the_urls_class` — `Urls` is in memory with autoload disabled after `init()`.
* `filter_http_request_args_uses_the_url_resolved_at_init` — the filter matches the URL cached at init, not one resolved at call time.

### DIFF
--- a/doc/plugin-upgrades.md
+++ b/doc/plugin-upgrades.md
@@ -1,0 +1,43 @@
+# Autoloading during a plugin upgrade
+
+Why [src/api/class-client.php](../src/api/class-client.php) resolves the API URL
+in `init()` instead of inside `filter_http_request_args()`.
+
+## The problem
+
+`Plugin_Upgrader` deletes the installed `speechkit/` directory and writes the new
+copy **within a request that already loaded the old version**. Our hooks are
+still registered, and `upgrader_process_complete` runs `wp_update_plugins()`,
+which issues an HTTP request — so `http_request_args` fires after the files
+backing our classes have gone.
+
+Any class the callback autoloads at that point is a fatal error, because
+Composer's classmap still points at the deleted path:
+
+```
+Warning: include(.../speechkit/vendor/composer/../../src/core/class-urls.php): Failed to open stream: No such file or directory
+Fatal error: Uncaught Error: Class "BeyondWords\Core\Urls" not found in .../src/api/class-client.php:93
+```
+
+The install itself has already succeeded by then, so the site recovers on the
+next request — but the user sees a white screen and gets a fatal-error email.
+
+It bites hardest on a **downgrade** (7.x → 6.x), where the replacement payload
+has an entirely different file layout and none of the 7.x paths resolve. A
+same-branch update only survives by luck: the new copy happens to contain a file
+at the same path.
+
+## The rule
+
+**Anything reachable from a hook that can fire during an upgrade must already be
+in memory.** In practice that means `http_request_args` (every outbound request,
+including core's update checks) and `admin_notices` (rendered on `update.php`).
+
+`Client::init()` runs during bootstrap, while the files are guaranteed to exist,
+so resolving `Urls::get_api_url()` there loads `Urls` for the whole request. That
+also covers `Settings::maybe_print_missing_creds_warning()`, which reaches for
+`Urls::get_dashboard_url()` from `admin_notices`.
+
+A fix can only ever protect *later* downgrades: the code running during the
+upgrade is the version being replaced, so a release can't repair the one before
+it.
--- a/readme.txt
+++ b/readme.txt
@@ -104,6 +104,15 @@ Yes. The plugin normally connects WordPress to BeyondWords through our REST API.
 
 == Changelog ==
 
+= 7.0.1 =
+
+Release date: TBC
+
+**Fixes**
+
+* [#621](https://github.com/beyondwords-io/wordpress-plugin/pull/621) Fix the fatal error shown when the plugin is replaced with a different version from the Plugins screen.
+    * Switching away from an existing 7.0.0 install still shows it once; the version being replaced has to be 7.0.1 or later for the fix to apply.
+
 = 7.0.0 =
 
 Release date: 12th August 2026
--- a/src/api/class-client.php
+++ b/src/api/class-client.php
@@ -67,11 +67,24 @@ class Client {
 	const VOICES_REQUEST_TIMEOUT = 8;
 
 	/**
+	 * API base URL, resolved once at bootstrap.
+	 *
+	 * @since 7.0.1
+	 *
+	 * @var string|null Null until `init()` runs.
+	 */
+	private static ?string $api_url = null;
+
+	/**
 	 * Register WordPress hooks.
 	 *
 	 * Must run early in the bootstrap so the filter precedes any API call.
 	 */
 	public static function init(): void {
+		// WordPress fires http_request_args mid-upgrade, after the plugin files have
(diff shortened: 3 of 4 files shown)


gouravkhunger (commented):
@galbus This is an automated PR review.

Soft score: **Approve**.
Minimal, correctly-targeted fix for a real fatal, with a test that proves the behaviour change; tidy up the two test-hygiene items (order-fragile `class_exists` assertion, reflection state not restored in `tearDown`) and the split API-URL source of truth inside `Client`.

<details>

## Architectural map

Single-point change in the API layer. `Client` gains a `private static ?string $api_url`, populated in `init()` (which runs at plugin-file load via `speechkit.php` → `Plugin::init()`), and `filter_http_request_args()` reads that instead of calling `Urls::get_api_url()` lazily. Side effect is the real fix: `Urls` is now resident in memory before `Plugin_Upgrader` can delete the file behind it.

Everything else is docs — a `= 7.0.1 =` changelog block and two PHPUnit tests.

## Affected subsystems

- `src/api/` — `Client` only. No other class touched.
- `readme.txt` changelog.
- `tests/phpunit/api/test-client.php`.

## Dependency graph changes

`Client` → `Urls` moves from lazy (call-time) to eager (bootstrap-time). No new edges, no removed ones. `Urls` has zero dependencies of its own — pure constants plus `defined()` guards — so eager-loading it can't cascade.

## Risky files

`src/api/class-client.php`. The filter runs on **every** outbound HTTP request site-wide, so a mistake here breaks core update checks, not just BeyondWords. The logic change is confined to one assignment and one read, and the `'' === $api_url` early-return still guards the null case.

## Public API changes

None. The new property is `private static`; `filter_http_request_args()` signature and return shape are unchanged. No hooks, filters, options, or REST routes added or removed.

## DB changes

None.

## Auth/security changes

Neutral, with one thing to be aware of. The `X-Api-Key` injection is now gated on a value frozen at bootstrap rather than resolved per call. The fail mode is fail-closed (no header) rather than leaking the key to a wrong host, which is the right direction. The `str_starts_with` prefix match is unchanged, so no new host-confusion surface.

## Deployment/config changes

None in code. `Stable tag:` still `7.0.0` — correct, version bumps happen at release.

CI: **Plugin Check fails**, but on an unrelated pre-existing issue — `Tested up to: 7.0 < 7.1`. That failure gates the Cypress job, which is reported as `skipping`, so no E2E ran on this branch. PHPUnit passes on 8.0 and 8.5; GrumPHP, ESLint, Jest pass.

## Findings

**1. Two sources of truth for the API URL inside `Client` (maintainability).** The filter uses `self::$api_url`; the ~15 request methods still call `\BeyondWords\Core\Urls::get_api_url()` inline (`class-client.php:151`, `171`, `272`, …). The values are identical in practice, so no bug today — but the next reader has to work out why one path is cached and the others aren't. A private `api_url()` accessor used everywhere would make the invariant self-evident.

**2. A late-defined `BEYONDWORDS_API_URL` now desyncs the filter from the request URL.** If the constant is defined after plugin load — a `plugins_loaded`/`init` hook rather than `wp-config.php` or an mu-plugin — the request methods build URLs against the new host while the filter still compares against the bootstrap value, and the auth header is silently dropped. Documented usage is `wp-config.php` (loads first) and the test bootstrap defines it before `muplugins_loaded`, so nothing in-repo hits this. Worth knowing it's now a constraint rather than a free choice.

**3. The eager-load approach is per-class, not per-surface.** The PR body states the general rule correctly, but enforcing it means remembering it at every future call site. A structurally complete alternative is one callback on `upgrader_process_complete` at priority 1 that unregisters our hooks when the target plugin is `speechkit` — that neutralises the entire hook surface (including the `admin_notices` path on `update.php`) in one place instead of pre-loading classes one at a time. The same limitation applies either way: only the version *doing* the upgrade is protected. The current surface checks out — `PostsList\Notices` uses only `self::`, and `Settings::maybe_print_missing_creds_warning()` needs `Settings\Utils` and `Urls`, both of which `Plugin::init()` now loads — so this is a durability suggestion, not a gap in today's code.

**4. `init_loads_the_urls_class` is order-fragile.** It asserts `class_exists( Urls::class, false )` in a shared process where `speechkit.php` and every other spec have already run. `phpunit.xml` sets `executionOrder="depends,defects"`, so a defects-first reshuffle can put a spec that touches `Urls` ahead of it, and the assertion passes regardless of the fix. `@runInSeparateProcess` — or dropping the test and relying on finding 5's stronger one — would make it mean something.

**5. Reflection state isn't restored on failure.** `filter_http_request_args_uses_the_url_resolved_at_init` overwrites the private static and restores it by calling `Client::init()` at the end of the test body. If an earlier assertion fails, `https://api.example.test/v1` leaks into every subsequent test in the process. Move the restore into `tearDown()`. This is the test that actually proves the fix, so it's worth hardening.

**Non-issues checked:** the typed static property is fine on the PHP 8.0 floor; `Client::init()` runs at file load, well before any upgrade can start, so the cache is always warm; `doc/plugin-upgrades.md` was added and removed within the branch, hence the net 3-file diff.

## Review priority order

1. `src/api/class-client.php:86` and `:105` — the actual fix. Correct for the reported trace.
2. `tests/phpunit/api/test-client.php` — findings 4 and 5.
3. CI — confirm the Plugin Check failure is the known `Tested up to` issue and not something this branch introduced. It's currently hiding Cypress.
4. `readme.txt` — the entry is in the right block under `**Fixes**`. The sub-bullet clears the bar; it's a user-actionable caveat about the 7.0.0 → 6.x case, not implementation detail.

## Verdict

Approve. Minimal, correctly-targeted fix for a real fatal, with a test that exercises the actual behaviour change. Findings 1, 4, and 5 are worth a follow-up commit; 2 and 3 are context for later, not blockers.

</details>

galbus:
@gouravkhunger a PR for the plugin version update is at https://github.com/beyondwords-io/wordpress-plugin/pull/624. This should be merged fist, then CI should pass here.

Edit: this is complete ✅

gouravkhunger (approved):

